PR #83, opened 2023-10-31
Read/Write Pools

prxt6529:


src/db-api.ts
```diff
@@ -108,12 +137,20 @@ export function execSQL(sql: string): Promise<any> {
   });
 }
 
-export function execSQLWithParams(
+export function execSQLWithParams(sql: string, params?: Record<string, any>) {
+  if (sql.startsWith('SELECT')) {
```
TarmoKalling, line 141: 1. `select * from mytable` would be sent to writer pool
2. `with my_temp_view as (select * from my_table) select * from my_view` would be sent to writer pool. We are not using "with" statements yet, but I've used them previously quite a lot. They are very useful with complex queries.

I would write this query other way around. If statement starts with insert, update, delete then use writer. In every other case use reader. This would quickly bring out cases where we wrongly use reader and would make sure we never use writer for reads.

src/db-api.ts
```diff
@@ -66,7 +76,19 @@ export async function connect() {
     process.exit();
   }
   port = +process.env.DB_PORT;
-  mysql_pool = mysql.createPool({
+  write_pool = mysql.createPool({
+    connectionLimit: 5,
+    connectTimeout: 30 * 1000,
```
TarmoKalling, line 81: This is just a suggestion, but we have `Time` class now. With this you could write this like: `Time.seconds(30).toMillis()` which IMO makes things more easily readable.

TarmoKalling (changes requested):

TarmoKalling (approved):

PR #92, opened 2023-11-03
Ability to force pool on concrete queries and its integration to profiles

TarmoKalling:


src/db-api.ts
```diff
@@ -143,9 +153,10 @@ export function execSQL(sql: string): Promise<any> {
 
 export function execSQLWithParams(
   sql: string,
-  params?: Record<string, any>
+  params?: Record<string, any>,
+  options?: { forcePool?: DbPoolName }
```
prxt6529, line 157: can we add `DbQueryOptions` here for type?

TarmoKalling: Good point. Changed it.

prxt6529 (approved):

PR #103, opened 2023-11-10
Web royalties

prxt6529:


src/transactionsReplayLoop/index.ts
```diff
@@ -1,17 +1,18 @@
...
-  const fromBlock = 16485000;
+  const fromBlock = 0;
   const toBlock = undefined;
   logger.info(`[RUNNING] [FROM BLOCK ${fromBlock}] [TO BLOCK ${toBlock}]`);
-  await loadEnv();
-  // await transactions(fromBlock, toBlock);
-  // await findOwnerMetrics();
-  await tdhLoop(true);
+  await loadEnv([Transaction]);
+  await transactions(fromBlock, toBlock);
```
TarmoKalling, line 15: Would be nicer if this functions name would be verb instead of a noun. A verb which would describe what it does. I as unfamiliar reader of this code don't know what it does. So something like `fetchAndPersistTransactions` maybe?

TarmoKalling (changes requested):

package.json
```diff
@@ -7,6 +7,7 @@
     "api:local": "NODE_ENV=local node src/api-serverless/dist/index.js",
     "backend:dev": "NODE_ENV=development node build/backend.js",
     "api:dev": "NODE_ENV=development node src/api-serverless/dist/index.js",
+    "backend:prod": "NODE_ENV=production node build/backend.js",
```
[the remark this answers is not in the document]
prxt6529: sometimes i need to call this in prod
eg when we replayed transactions - i created a new instance of backend on an ec2 and used the above to replay all transactions since it would be a long task taking much longer than lambda lifespan

src/transaction_values.ts
```diff
@@ -83,113 +110,166 @@ async function resolveValue(t: Transaction, receipt: any, events: number) {
...
+      contract,
+      tokenId,
+      royaltiesAmount,
+      totalAmount
+    };
+  }
+};
+
+// HELPER FUNCTION FOR DEBUGGING VALUES USING TRX HASHES FROM DB
+export const debugValues = async () => {
```
prxt6529, line 229: @TarmoKalling 
this is just a debug version which i use to resolve transaction values when needed - i renamed and addd a comment

about the 'transactions' array further down - it's just a collection of sample transactions 

tbh this should be a test....

src/transaction_values.ts
```diff
@@ -83,113 +110,166 @@ async function resolveValue(t: Transaction, receipt: any, events: number) {
...
+const parseSeaportLog = async (
+  t: Transaction,
+  royaltiesAddress: string,
+  log: ethers.providers.Log
+) => {
+  let seaResult;
+  try {
+    seaResult = SEAPORT_IFACE.parseLog(log);
+  } catch (err: any) {
+    // logger.error('SEAPORT PARSE ERROR', t.transaction, err);
```
TarmoKalling, line 170: You can change it to:
```
logger.debug(`SEAPORT PARSE ERROR for transaction ${t.transactio}`, err);
```
This way it's usually not logged but you can activate this log without changing code.

TarmoKalling (approved):

PR #104, opened 2023-11-13
API logging with tracing and levels using Winston framework

TarmoKalling:


src/votes.ts
```diff
@@ -11,6 +11,9 @@ import { VoteEvent, VoteEventReason } from './entities/IVoteEvent';
 import { Time } from './time';
 import { BadRequestException } from './exceptions';
 import * as tdh_consolidation from './tdh_consolidation';
+import { Logger } from './logging';
+
+const logger = Logger.get('VOTES');
```
prxt6529, line 16: do we want to separate this from the previous Logger.get('VOTES'); in votes.routes?

TarmoKalling: Done

prxt6529 (approved):

PR #137, opened 2023-11-21
Sonarcloud 20/11

prxt6529:


src/db.ts
```diff
@@ -352,56 +313,50 @@ export async function fetchLatestDelegationsBlockNumber() {
...
-}
-
-export async function fetchTdhReplayTimestamp(date: Date) {
-  const sql = `SELECT timestamp FROM ${TDH_BLOCKS_TABLE} WHERE created_at > '2023-03-06' AND timestamp <= ${mysql.escape(
-    date
-  )} order by block_number asc limit 1;`;
-  const r = await execSQL(sql);
-  return r.length > 0 ? r[0].timestamp : null;
+  const r = await sqlExecutor.execute(sql);
+  return r.length > 0 ? new Date(r[0].timestamp) : new Date();
```
TarmoKalling, line 329: Are you sure that in case there is nothing, the timestamp should be current time and not epoch for example?

src/db.ts
```diff
@@ -432,98 +390,56 @@ export async function fetchAllMemeLabNFTs(orderBy?: string) {
...
 export async function fetchConsolidationDisplay(
   myWallets: string[]
 ): Promise<string> {
-  const sql = `SELECT * FROM ${ENS_TABLE} WHERE wallet in (${mysql.escape(
-    myWallets
-  )})`;
-  const results = await execSQL(sql);
+  const sql = `SELECT * FROM ${ENS_TABLE} WHERE wallet IN (:wallets)`;
+  const results = await sqlExecutor.execute(sql, {
+    wallets: myWallets.join(',')
```
TarmoKalling, line 430: I think you dont have to do this join thing. sqlExecutor takes care of this when you pass just an array. Look at db.ts:201

src/tdh.ts
```diff
@@ -168,7 +168,7 @@ export const findTDH = async (lastTDHCalc: Date) => {
         areEqualAddresses(t.contract, GRADIENT_CONTRACT)
       );
 
-      ADJUSTED_NFTS.map((nft) => {
+      ADJUSTED_NFTS.forEach((nft) => {
         let tokenTransactions: Transaction[] = [];
         if (areEqualAddresses(nft.contract, MEMES_CONTRACT)) {
           tokenTransactions = [...memesTransactions].filter(
```
TarmoKalling, line 174: just fyi. this seems like a pointless loop since tokenTransactions is an unused variable

src/db.ts
prxt6529: will be fixing this one good catch!

src/db.ts
prxt6529: you're right

src/tdh.ts
prxt6529: loop isn't pointless, just the declaration and assigning of 'tokenTransactions'

src/api-serverless/src/app.ts
```diff
@@ -360,27 +360,24 @@ loadApi().then(() => {
...
-        returnPaginatedResult(result, req, res);
-      }
-    );
+    db.fetchLabNFTs(memeIds, pageSize, page, nfts, sortDir).then((result) => {
+      result.data.map((d: any) => {
+        d.meme_references = JSON.parse(d.meme_references);
+        d.metadata = JSON.parse(d.metadata);
+        if (
+          d.metadata.animation_details &&
+          typeof d.metadata.animation_details === 'string'
```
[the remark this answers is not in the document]
prxt6529: not sure tbh - but thats how it was so don't really wanna change functionality

TarmoKalling (approved):

PR #155, opened 2023-11-27
Block user from giving CIC ratings to themselves

TarmoKalling:


src/rates/rates.service.ts
```diff
@@ -78,6 +78,9 @@ export class RatesService {
     if (amount === 0) {
       return;
     }
+    if (matter === 'CIC' && raterProfileId === matterTargetId) {
+      throw new BadRequestException('User cannot rate itself');
```
prxt6529, line 82: Might be better to say "Users cannot rate themselves"

prxt6529 (approved):

src/rates/rates.service.ts
TarmoKalling: Fixed

PR #305, opened 2024-02-08
Nextgen - New Implementation for Uploader and Missing Resolutions

prxt6529:


src/nextgen/nextgen_generator.ts
```diff
@@ -11,6 +23,46 @@ export interface Details {
...
+    const response = await s3.send(command);
+    response.Contents?.forEach((object) => {
+      if (object.Key) {
+        contents.push(parseInt(object.Key?.replace(path, '')));
+      }
+    });
+  } catch (error) {
+    logger.error(`[S3 LIST ERROR] : [PATH ${path}] : [ERROR ${error}]`);
+  }
+  return contents;
```
TarmoKalling, line 45: If an error is caught on line 43, it should return an empty array? Wouldn't we rather want the loop to crash and error propagated to Sentry?

src/nextgenMissingImageResolutions/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  s3 = new S3Client({ region: NEXTGEN_BUCKET_AWS_REGION });
+  cloudfront = new CloudFrontClient({ region: NEXTGEN_BUCKET_AWS_REGION });
+}
+
+const START_INDEX = 10000000000;
+const END_INDEX = 10000000999;
+
+const BATCH_SIZE = 15;
+
+export const handler = async (event: any) => {
```
TarmoKalling: event is unused variable

src/nextgenMissingImageResolutions/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  const generatorPath = `${GENERATOR_BASE_PATH}/mainnet/png/${tokenId}/${resolution}`;
+  const s3Path = `mainnet/png${resolution}/${tokenId}`;
+
+  logger.info(
+    `[TOKEN_ID ${tokenId}] : [RESOLUTION ${resolution.toUpperCase()}] : [GENERATOR PATH ${generatorPath}] : [S3 PATH ${s3Path}]`
+  );
+
+  const imageBlob = await getImageBlobFromGenerator(generatorPath);
+  if (!imageBlob) {
+    logger.info(`[IMAGE BLOB ERROR] : [EXITING]`);
```
TarmoKalling: You are catching errors in getImageBlobFromGenerator and info logging them here. Why not just propagate the error up to the top so we could see it in Sentry?

src/nextgenMissingImageResolutions/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  logger.info(
+    `[UPLOADING BATCH] : [RESOLUTION: ${resolution.toUpperCase()}] : [BATCH ${JSON.stringify(
+      batch
+    )}]`
+  );
+  const promises = [];
+  for (let item of batch) {
+    promises.push(uploadMissingNextgenImage(item, resolution));
+  }
+  await Promise.all(promises);
```
TarmoKalling: I personally prefer this oneliner, but matter of taste:

```
  await Promise.all(
    batch.map((item) => uploadMissingNextgenImage(item, resolution))
  );
```

prxt6529: changed, much better 👍

src/nextgenMissingImageResolutions/index.ts
prxt6529: i change to logger.error 😬 

i expect that this will fail a lot - and by a lot i mean 5-10% of requests so i think i'll be spamming
its not an error that this logic made, its another service failing to give it a response
and i want the batch to carry on

src/nextgen/nextgen_generator.ts
prxt6529: contents is empty at this point isn't it?

TarmoKalling (approved):

PR #339, opened 2024-02-22
Contract agnostic transactions loop

TarmoKalling:


src/transactionsLoopV2/serverless.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+functions:
+  memesTransactionsLoop:
+    handler: index.handler
+    name: memesTransactionsLoop
+    description: ${env:VERSION_DESCRIPTION}
+    role: arn:aws:iam::987989283142:role/lambda-vpc-role
+    reservedConcurrency: 1
+    events:
+      - schedule: cron(1 0 * * ? *)
```
prxt6529, line 23: is this schedule right? is it every minute?
why not use the more readable `rate(x minutes)`

prxt6529 (commented):
Part of former transactions also does ens discovery - not sure if you accounted for this.

Also as a general comment I'm not sure i understand what the requirement you are solving here is? Is the requirement 'rewrite transactions loop' ?

src/transactionsLoopV2/serverless.yaml
TarmoKalling: You are right. It is not correct. I thought I copied this from transactionsLoop, but it was from tdhLoop. Fixed it.

TarmoKalling:
@prxt6529 (Github doesn't let me respond to general comments directly)
To answer your questions:
1. Good catch. Added the ENS discovery.
2. My requirement is/was not to rewrite transactionsLoop. My big requirement is to have quick and easy to understand queries. To reach that goal we need to get rid of collection specific database tables. To get rid of those, it makes sense to get rid of collection specific lambdas (at least for generic flows like finding transactions/owners/nfts etc. I don't want there to be special code written in special time for a special collection. So my goal with this PR was to create a contract agnostic version of transactions loop where there is no half-saved blocks and which could be deployed for new contracts "with one click". My secondary goal was to reduce technical dept by refactoring. I started off by trying to get original transactionsLoop to do what I wanted, but it was real hard for me to understand how this code flows. Where to put the transactions in order to make the new start from (lastblock+1) work for example. So I started writing my own structure for this. And in the end phase I discovered that this can double as transactionsReplayLoop also, so we would get rid of this too. So yea, there was a lot of rewriting, but I didn't do it for the sake of rewriting :)

prxt6529:
i would think that the current transactions.ts file would be easy to use - it already accepts contract

```
export const findTransactions = async (
  startingBlock: number,
  latestBlock?: number,
  pageKey?: string,
  contracts?: string[]
  ```
  so i would think that the only changes that would be needed is making this contracts input maybe not an array and not optional
  
  the fundamental difference is that this is written in a way that it picks 'end block' at the beginning and runs until that block
  
  which i think is more reusable in terms of replaying transactions from 'from_block' to 'to_block'
  
  if im not mistaken - your implemantation would rerun transaction from 'from_block' to now

TarmoKalling:
> if im not mistaken - your implemantation would rerun transaction from 'from_block' to now

True.. I can add to_block and add a deployment version of transactionReplayLoop which would take this as env (otherwise it will run until "now".

Theoretically yes, it could have been done the way you said. Yesterday you said that there is an efficiency with the current way of doing things. It always redoes the last block. Even if it is already done. So if the last block would have 1000 transactions we would process those 1000 transactions every two seconds (alchemy calls and everything). To get rid of this behaviour was also goal of my version and so I did. That's why I need to always be sure I save full blocks (not so that half of the block is on the next page which might fail).

prxt6529:
but the only reason you need to implement the logic of "completed" block is because you are starting off with unknown "to_block" 

whereas if you started with alchemy's getBlockNumber() it would return the latest completed block so you don't need to worry about it 

also the inefficiency was a simple `+1` on the existing start block number

TarmoKalling:
> also the inefficiency was a simple +1 on the existing start block number

No, but this discussion moved to Discord.

prxt6529 (approved):

PR #340, opened 2024-02-23
Add support for nextgen 0.5k and thumbnail resolutions

prxt6529:


src/nextgenMediaImageResolutions/index.ts
```diff
@@ -13,32 +13,46 @@ import {
...
   await loadEnv([]);
   setup();
-  const resolutions = ['2k', '4k', '8k'];
-  for (let resolution of resolutions) {
-    const isFinished = await findMissingImages(resolution);
+
+  const resolutions = [Resolution['thumbnail'], Resolution['0.5k']];
+
+  for (let resolutionEnum of resolutions) {
+    const resolution = Resolution[resolutionEnum].toString();
```
TarmoKalling, line 50: Why convert it to string here? Why not pass it as a "Resolution" type to other functions?

TarmoKalling (approved):

src/nextgenMediaImageResolutions/index.ts
```diff
@@ -13,32 +13,46 @@ import {
...
-const BATCH_SIZE = 30;
+const BATCH_SIZE = 60;
 
 export const handler = async () => {
   const start = Time.now();
   logger.info(`[RUNNING]`);
   await loadEnv([]);
   setup();
-  const resolutions = ['2k', '4k', '8k'];
-  for (let resolution of resolutions) {
```
TarmoKalling, line 40: use const instead of let

a2rocket (approved):

src/nextgenMediaImageResolutions/index.ts
TarmoKalling: Only reason why you need it as string in certain places is to uppercase it for log/discord messages. 
1. I don't see why it has to be uppercased for those. 
2. Strong narrow types are always better, it eliminates the possibility of someone using those functions with unintended values.

PR #480, opened 2024-04-26
On-Chain Primary Address

prxt6529:


GelatoGenesis (approved):

src/profiles/profiles.db.ts
```diff
@@ -528,6 +524,16 @@ export class ProfilesDb extends LazyDbAccessCompatibleService {
     );
   }
 
+  async updatePrimaryAddress(param: {
+    profileId: string;
+    primaryAddress: string;
+  }) {
+    await this.db.execute(
+      `update ${PROFILES_TABLE} set primary_wallet = :primaryAddress where external_id = :profileId`,
```
brookr, line 532: Use ALLCAPS for the SQL keywords, as per tradition?

brookr: hm, seems we are mixed on this convention across the codebase. your thoughts, @prxt6529? Clean it up we go, or don't worry about it now?

src/profiles/profiles.service.ts
```diff
@@ -1009,6 +983,60 @@ export class ProfilesService {
...
+    }
+
+    const highestTdhAddress = await getHighestTdhAddressForConsolidationKey(
+      consolidationKey
+    );
+    if (highestTdhAddress) {
+      return highestTdhAddress;
+    }
+
+    return wallets[0].wallet.address;
```
brookr, line 1027: if no primary is set via delegation, and no wallet has highest tdh, then it just returns the first wallet in the list? 

is the list already sorted by any criteria? is it date-added, or alphabetic?

PR #496, opened 2024-05-02
Proxy received granted map profile min

analyticsflowee:


src/api-serverless/src/proxies/proxy.api.service.ts
```diff
@@ -199,7 +207,48 @@ export class ProfileProxyApiService {
...
+        sort_direction
+      }),
+      this.profileProxiesDb.countProfileGrantedProfileProxies({
+        created_by
+      })
+    ]);
+    return {
+      count,
+      page: page,
+      next: profileProxies.length === page_size,
```
GelatoGenesis, line 246: this should be done by count. if you do it like this, you could send a false flag, if last page is exactly page size

src/api-serverless/src/generated/models/ProfileProxy.ts
```diff
@@ -10,13 +10,14 @@
  * Do not edit the class manually.
  */
 
+import { ProfileMin } from '../models/ProfileMin';
 import { HttpFile } from '../http/http';
 
 export class ProfileProxy {
     'id': string;
-    'target_id': string;
+    'granted_to'?: ProfileMin;
```
GelatoGenesis, line 18: why is this optional?

GelatoGenesis (changes requested):

src/api-serverless/src/profiles/proxies/profile-proxies.routes.ts
```diff
@@ -56,24 +56,64 @@ router.get(
...
+    >,
+
+    res: Response<ApiResponse<Page<ProfileProxy>>>
+  ) => {
+    const targetProfile =
+      await profilesService.getProfileAndConsolidationsByHandleOrEnsOrIdOrWalletAddress(
+        req.params.handleOrWallet
+      );
+    if (!targetProfile?.profile) {
+      throw new BadRequestException(
```
GelatoGenesis, line 77: This should be NotFoundException

GelatoGenesis (changes requested):

PR #512, opened 2024-05-09
fixed proxy action update params bug

analyticsflowee:


src/profile-proxies/profile-proxies.db.ts
```diff
@@ -334,6 +334,29 @@ export class ProfileProxiesDb extends LazyDbAccessCompatibleService {
     );
   }
 
+  private getUpdateProfileProxyActionQuery({
```
GelatoGenesis, line 337: Not a fan of refactoring query composition into another function. Makes it harder to read the code IMO.

GelatoGenesis (approved):

PR #586, opened 2024-06-25
Prenode Service Registration

prxt6529:


src/api-serverless/src/oracle/api.oracle.routes.ts
```diff
@@ -306,3 +309,29 @@ router.get(
...
+  if (domain === undefined) {
+    return res.status(400).send({ message: 'Missing domain' });
+  }
+  if (!tdh) {
+    return res.status(400).send({ message: 'Missing tdh' });
+  }
+  if (!block) {
+    return res.status(400).send({ message: 'Missing block' });
+  }
+  // if (!isValidIP(ip)) {
```
brookr, line 330: Can delete?

brookr (changes requested):
looks good, just need to deal with the commented-out isValidIP function.

src/api-serverless/src/oracle/api.oracle.routes.ts
```diff
@@ -306,3 +309,29 @@ router.get(
...
+  }
+  if (!block) {
+    return res.status(400).send({ message: 'Missing block' });
+  }
+  if (!isValidIP(ip) && !isLocalhost(ip)) {
+    return res.status(400).send('Invalid IP');
+  }
+
+  const validation = await db.validatePrenode(ip, domain, tdh, block);
+  const statusCode = validation.block_sync && validation.tdh_sync ? 200 : 400;
```
brookr, line 335: this will send a 400 (when out of sync), even if the registration is created? i think `200` for successful registration makes more sense, regardless of the sync status. eg: the prenode was registered, so 200 OK. I like keeping 400 to indicate something improper in the formating of the request, like the above checks.

brookr (changes requested):
1 question on status codes

src/api-serverless/src/oracle/api.oracle.routes.ts
prxt6529: i been thinking about this too... i agree it shouldn't be 400 for invalid
i keep going back and forth between 200 for "ok" when invalid or 201?
and maybe 204 for "created but invalid"?

brookr: 201 - created 
the prenode might sync later, so i don't think the status code needs to report on that.

brookr (approved):
👍👍

PR #643, opened 2024-07-23
All loops to use auto-closing db connections and have auto-timing

GelatoGenesis:


src/secrets.ts
```diff
@@ -1,11 +1,29 @@
...
+): Promise<T> {
+  const start = Time.now();
+  const logger = opts?.logger ?? Logger.get('MAIN');
+  logger.info(`[RUNNING]`);
+  await loadEnv(opts?.entities ?? []);
+  try {
+    return await fn();
+  } finally {
+    logger.info(`[FINISHED IN ${start.diffFromNow().formatAsDuration()}]`);
+    await unload();
```
prxt6529, line 23: maybe unload should be before the log line?

GelatoGenesis: Debatable, but I personally think it fits better here because unloading takes hardly any time but can theoretically fail by which point the loop itself has finished it's business logic and therefor should log that it's finished.

Anyway it's so obscure. Since I already deployed all the loops to staging and it's so debatable I'd leave it as is.

prxt6529 (approved):

PR #767, opened 2024-10-07
Full Redis integration for API

GelatoGenesis:


src/redis.ts
```diff
@@ -75,10 +72,3 @@ export async function initRedis() {
   await redis.connect();
   logger.info('finished connecting');
 }
-
-export async function disconnectRedis() {
-  if (redis) {
-    await redis.disconnect();
-    logger.info('Disconnected!');
-  }
-}
```
prxt6529, line 84: no need for disconnect?

GelatoGenesis: i experimented with it, but the fact is if lambda is killed the killing happens quicker that this code is reached. i tried with all kinds of strategies found from interwebs, but killing always happened quicker

prxt6529 (approved):

PR #878, opened 2025-03-07
Wave notification settings

prxt6529:


src/pushNotificationsHandler/sendPushNotifications.ts
```diff
@@ -36,12 +36,15 @@ export async function sendMessage(
...
   notification_id: number,
   extra_data: any,
+  badge?: number,
   imageUrl?: string
 ) {
   init();
   title = title.replace(/@\[(.+?)\]/g, '@$1');
   body = body.replace(/@\[(.+?)\]/g, '@$1');
 
+  const badgeNumber = Number.isFinite(Number(badge)) ? Number(badge) : 1;
```
GelatoGenesis, line 46: Why would it be infinite? If it's undefined or smth? If so then perhaps a cleaner version would be `const badgeNumber = parseIntOrNull(badge) ?? 1`

PR #881, opened 2025-03-12
Websockets framework which works both in lambdas and locally with authentication and client tracking

GelatoGenesis:


src/api-serverless/src/app.ts
```diff
@@ -610,6 +622,14 @@ loadApi().then(async () => {
     });
   });
 
+  apiRouter.get(`/test`, async function (req: any, res: any) {
```
GelatoGenesis, line 625: I'll leave this here for testing for now unitl we get some real integration

PR #892, opened 2025-03-21
Direct Messages

prxt6529:


src/api-serverless/openapi.yaml
```diff
@@ -2245,6 +2245,32 @@ paths:
                 $ref: "#/components/schemas/ApiWave"
         "400":
           description: Invalid request
+  /waves/create-direct-message:
```
GelatoGenesis, line 2248: Doesn't this conflict with POST /waves/{id} ?

src/profiles/profiles.db.ts
```diff
@@ -219,6 +219,18 @@ export class ProfilesDb extends LazyDbAccessCompatibleService {
       );
   }
 
+  public async getHandlesByAddresses(
+    addresses: string[],
+    connection?: ConnectionWrapper<any>
+  ): Promise<string[]> {
+    const result = await this.db.execute(
+      `select handle from ${PROFILES_TABLE} where primary_wallet in (:addresses)`,
```
GelatoGenesis, line 227: Not sure what you are trying to do here, but are you not interested in cases where input is a non-primary address?

GelatoGenesis: If not then your function name should be more specific. Otherwise someone might assume that it's "give any wallet, get handle" thing

PR #1003, opened 2025-05-18
Implement dual-precision TTL for redisCached

punk6529:
## Summary
- add `TtlUnit` to specify seconds vs milliseconds
- extend `redisCached` with TTL unit argument and overload
- support corrupt-cache eviction and falsy cache entries
- add tests for EX, PX, falsy payloads, and corrupt JSON

## Testing
- `npm test` *(fails: jest not found)*
- `npx tsc --noEmit` *(fails: missing type definitions)*

### DIFF
--- a/src/redis.ts
+++ b/src/redis.ts
@@ -5,6 +5,8 @@ import { Time } from './time';
 
 let redis: Redis;
 
+export type TtlUnit = 'seconds' | 'milliseconds';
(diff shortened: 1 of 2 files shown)


GelatoGenesis:
this code is subject to deletion anyway as nothing uses it, so i'll close this pr

PR #1006, opened 2025-05-18
Add ETH price sync test coverage

punk6529:
## Summary
- implement python module `eth_price_sync` with logic similar to ETH price loop
- add unit tests covering historic and current price paths

## Testing
- `pytest -q` *(fails: command not found)*


punk6529:
in error!

PR #1038, opened 2025-05-27
Drop Reactions

prxt6529:


package.json
```diff
@@ -108,6 +110,7 @@
   },
   "overrides": {
     "elliptic": "6.6.1",
-    "ws": "8.18.0"
+    "ws": "8.18.0",
+    "undici": "7.10.0"
```
GelatoGenesis, line 114: what does this do?

src/api-serverless/openapi.yaml
```diff
@@ -4041,6 +4043,14 @@ components:
...
         is_signed:
           type: boolean
+        reactions:
+          type: array
+          items:
+            $ref: "#/components/schemas/ApiDropReaction"
+        context_profile_reaction:
+          anyOf:
+            - type: "null"
+            - type: string
```
GelatoGenesis, line 4053: 1. This should be under `context_profile_context` just as `reaction`
2. Try to avoid anyOf null/aThing at any cost because it's mot that well supported by swagger. Only case you really need to use it is when you reference a complex type or null. In simple types it should be describes as:
```
     your_field:
          type: string
          nullable: true
```

src/api-serverless/openapi.yaml
```diff
@@ -4158,6 +4168,18 @@ components:
...
+      required:
+        - reaction
+        - profiles
+      properties:
+        reaction:
+          type: string
+        profiles:
+          type: array
+          items:
+            $ref: "#/components/schemas/ApiProfilePreview"
```
GelatoGenesis, line 4182: 1. Why not use ApiIdentity instead? We have all mechanisms in place already to easily pull those from DB very efficiently and we already use it everywhere else.
2. ApiProfilePreview.handle shouldn't be nullable. If you reacted, you must have had a profile. If you have a profile you must have a handle.

src/api-serverless/openapi.yaml
```diff
@@ -4373,6 +4397,14 @@ components:
             $ref: "#/components/schemas/ApiDropSubscriptionTargetAction"
         is_signed:
           type: boolean
+        reactions:
+          type: array
+          items:
+            $ref: "#/components/schemas/ApiDropReaction"
+        context_profile_reaction:
```
GelatoGenesis, line 4404: Here too. Put it under context_profile_context.

src/api-serverless/package.json
```diff
@@ -68,5 +68,8 @@
     "typescript": "^5.0.4"
   },
   "author": "The serverless webpack authors (https://github.com/elastic-coders/serverless-webpack)",
-  "license": "MIT"
+  "license": "MIT",
+  "overrides": {
+    "jsonpath-plus": "10.3.0"
```
GelatoGenesis, line 73: why do we need jsonpath-plus?

src/api-serverless/src/drops/drop.validator.ts
```diff
@@ -20,6 +20,15 @@ export const ApiAddRatingToDropRequestSchema: Joi.ObjectSchema<ApiAddRatingToDro
     category: Joi.string().optional().allow(null) // for legacy reasons
   });
 
+export interface ApiAddReactionToDropRequest {
```
GelatoGenesis, line 23: This should be described in openapi.yaml. Everything that starts with Api should be described there.

src/api-serverless/src/drops/reactions/drop-reactions.service.ts
```diff
@@ -0,0 +1,142 @@
...
+    private readonly dropsService: DropsApiService
+  ) {}
+
+  public reactionHandler =
+    (isDeleting: boolean): RequestHandler =>
+    async (
+      req: Request<any, any, ApiAddReactionToDropRequest, any, any>,
+      res: Response<ApiResponse<ApiDrop>>,
+      next: NextFunction
+    ) => {
```
GelatoGenesis, line 39: What's the reasoning for using this pattern where routes stuff is in service as a "thingHandler"? Not sure how it would help us... We have drop-reactions.service which is half-routes, half-service and we have reaction.service which is full service. Seems a bit confusing. Why not do it like we've always done? API specific stuff in routes file, one service file, one db file?

src/api-serverless/src/drops/reactions/reactions.db.ts
```diff
@@ -0,0 +1,152 @@
...
+    const {
+      profileId,
+      dropId,
+      waveId,
+      reaction: reactionStr,
+      isDeleting
+    } = payload;
+
+    if (isDeleting) {
+      // ----- DELETE PATH -----
```
GelatoGenesis, line 34: Useless comments probably left from AI.

GelatoGenesis (changes requested):

src/api-serverless/openapi.yaml
```diff
@@ -4041,6 +4043,14 @@ components:
             $ref: "#/components/schemas/ApiDropSubscriptionTargetAction"
         is_signed:
           type: boolean
+        reactions:
```
GelatoGenesis, line 4046: Also you haven't descrived the reacting endpoint in openapi.yaml

PR #1147, opened 2025-10-16
TDH Editions

prxt6529:


src/api-serverless/openapi.yaml
```diff
@@ -2227,6 +2227,223 @@ paths:
...
+      responses:
+        "200":
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ApiTdhEditionsPage"
+        "400":
+          description: Invalid wallet
+  /tdh-editions/consolidation/{consolidation_key}:
```
GelatoGenesis: This is redundant IMO. Because /tdh-editions/identity/{identity} already does that. If you don't know the handle then you just do /tdh-editions/identity/{any-random-wallet-from-consolidation-key}.

src/api-serverless/openapi.yaml
```diff
@@ -5935,6 +6152,59 @@ components:
...
+        balance:
+          type: integer
+          format: int64
+        hodl_rate:
+          type: integer
+          format: int64
+        days_held:
+          type: integer
+          format: int64
+        wallet:
```
GelatoGenesis: Instead of all the fields down from here, why not just return ApiProfileMin like all other endpoints do?

src/api-serverless/src/tdh-editions/tdh-editions.routes.ts
```diff
@@ -0,0 +1,136 @@
...
+});
+
+router.get('/identity/:identity', async (req, res) => {
+  const identityParam = req.params.identity;
+  const timer = Timer.getFromRequest(req);
+  const filterType = UUID_REGEX.test(identityParam)
+    ? IdentityFilterType.PROFILE_ID
+    : IdentityFilterType.HANDLE;
+
+  const identityRecord =
```
GelatoGenesis: if you would get rid of the consolidation alternative, you could do this here: `identityFetcher.getIdentityAndConsolidationsByIdentityKey(wallet_or_profile_id_or_handle_or_normalised_handle_or_ens)`

GelatoGenesis (changes requested):

src/db.ts
```diff
@@ -745,17 +749,27 @@ export async function persistTDH(
               wallet: wallet.toLowerCase()
             })
             .execute();
+          await tdhEditionsRepo
+            .createQueryBuilder()
+            .delete()
+            .where('LOWER(wallet) = :wallet ', {
```
GelatoGenesis: Why not insert wallets to tdh_editions table already at lowercase? Would make the queries indexable.

src/db.ts
```diff
@@ -811,17 +827,27 @@ export async function persistConsolidatedTDH(
               walletPattern
             })
             .execute();
+          await tdhEditionsRepo
+            .createQueryBuilder()
+            .delete()
+            .where('consolidation_key like :walletPattern', {
```
GelatoGenesis: why? you have a concrete column 'wallet' there. you could do an indexable exact equals.

PR #1179, opened 2025-11-14
Proxy credit spend taken out of transaction

GelatoGenesis:


prxt6529 (changes requested):
It looks ok for api side but are you 100% sure this is ok to do? 
I see in frontend that responses expect and use fields like 'total_cic_rating' and 'cic_rating_by_user' which are no longer being returned

PR #1494, opened 2026-04-07
Fix TDH full-set boost for current season

prxt6529:
## Summary by CodeRabbit

* **Refactor**
  * Improved boost calculations for card sets and seasons: more consistent season eligibility, higher numeric precision, and rounded display values. Boost breakdowns and available/acquired info strings are now rebuilt to reflect the refined computations.

* **Tests**
  * Updated tests to compute expected boost values dynamically and added assertions ensuring breakdowns and info strings match displayed results.


src/tdhLoop/tdh.ts
```diff
@@ -41,13 +41,53 @@ const logger = Logger.get('TDH');
...
+export const ADDITIONAL_CARD_SET_BOOST = 0.05;
+export const ADDITIONAL_CARD_SET_RATIO = 0.6529;
+
+function roundBoostValue(value: number): number {
+  return Math.round(value * 1e6) / 1e6;
+}
+
+function getBoostableSeasons(seasons: MemesSeason[]): MemesSeason[] {
+  const maxSeasonId =
+    seasons.length > 0 ? Math.max(...seasons.map((s) => s.id)) : 0;
```
GelatoGenesis, line 54: Math.max([]) is 0 so that shorthand if here is kinda pointless imo

src/tests/calculate-boost.test.ts
```diff
@@ -26,6 +30,31 @@ const seasons: MemesSeason[] = seasonData.map((data, index) => ({
   boost: 0.05
 }));
 
+const roundToTwoDecimals = (value: number) => Math.round(value * 100) / 100;
+
+const roundToSixDecimals = (value: number) => Math.round(value * 1e6) / 1e6;
```
GelatoGenesis: we have numbers.ts. Perhaps create a function there? Seems like a useful reusable thing.

```
  public roundDecimals(original: number, decimals: number): number {
    const exp = Math.pow(10, decimals);
    return Math.round(original * exp) / exp;
  }
```

src/tdhLoop/tdh.ts
prxt6529: it would be `Math.max(...[])`
which comes out to -Infinity

GelatoGenesis: But why do you even do the spread there?

prxt6529: because its an array 👀 
image

GelatoGenesis: damn. i just tested with an empty array, and it worked. but yes, it doesn't work with non-empty array. i did not know that. wonders of javascript. another way to solve the whole function with oneliner though:

```
seasons.sort((a, d) => d.id - a.id).slice(1).filter((it) => it.boost)
```

but i approved the pr

GelatoGenesis (approved):

PR #1636, opened 2026-06-16
[codex] Harden v2 wave feed contracts

punk6529:
## Issue
- The V2 wave drops contract was too loose: docs implied one replacement path, while handler behavior silently normalized or ignored invalid query params.
- That made it unclear whether `/v2/waves/{id}/drops` was the wave timeline feed, a reply-thread feed, a curation feed, or all of the above.

## Fix
- Make `/v2/waves/{id}/drops` explicitly documented and enforced as the V2 wave timeline feed.
- Add `GET /v2/drops/{id}/replies` as the dedicated V2 reply-thread route.
- Update V1 deprecation guidance to point clients to the correct V2 route for timeline, replies, and curation drops.
- Preserve existing timeline semantics and the documented max `limit` of 200 for compatibility.

## Changes
- Added Joi validation for shared V2 feed query params with `limit` default 50, max 200, positive `serial_no_limit`, valid `search_strategy`, and valid `drop_type`.
- Reject unsupported `drop_id` and `curation_id` on `/v2/waves/{id}/drops` so docs and runtime behavior match.
- Added V2 reply-feed service flow that resolves the parent drop's owning visible wave before returning the reply feed.
- Regenerated OpenAPI route artifacts for the new route.
- Documented the V2 timeline/replies/curation route split in `docs/architecture.md`.
- Added handler and service tests for defaults, max limit, rejected invalid params, rejected legacy filters, and reply-feed resolution.

## Validation
- `npm run restructure-openapi`
- `npm test -- src/api-serverless/src/waves/waves-v2-handlers.test.ts src/api-serverless/src/waves/api-wave-v2-service.test.ts src/drops/drops.db.test.ts src/api-serverless/generate-openapi-routes.test.ts`
- `npm test -- src/api-serverless/src/waves/waves-v2-handlers.test.ts src/api-serverless/src/waves/api-wave-v2-service.test.ts`
- Manual Windows equivalent of `npm run lint`: `npx eslint "src/**/*.ts" --fix --max-warnings=0` with `NODE_OPTIONS=--max-old-space-size=8192`
- `npx tsc -p tsconfig.json --noEmit`
- Frontend compatibility search: current frontend does not send `drop_id` or `curation_id` to `/v2/waves/{id}/drops`; current limits are <= 50; reply UI still uses the older parent-drop query route.

## Risk
- Medium: this tightens API contract behavior for `/v2/waves/{id}/drops`. Clients sending undocumented `drop_id` or `curation_id` there now get a 400 instead of an ignored or misleading timeline response.
- Rollback: revert this PR; V1 routes are unchanged.

## Review Notes
- Please focus review on whether the new V2 route split is the right long-term contract before we port more V1 read behavior.

## Summary by CodeRabbit

* **New Features**
  * Added a new V2 API endpoint to fetch a drop’s reply thread, including pagination and optional filtering.
* **Bug Fixes**
  * Improved V2 wave feed query handling by consistently validating and normalizing pagination/filter parameters.
* **Documentation**
  * Updated architecture notes to clarify how v2 read boundaries split timeline, reply, and curation feeds.
  * Refreshed endpoint docs to distinguish timeline drops from reply-thread and curation drops, and to clarify pagination/response details.
* **Tests**
  * Added/extended coverage for reply-thread retrieval and feed query validation.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -214,6 +214,8 @@ Important API responsibilities:
 
 Wave rows can be top-level waves or subwaves through the nullable `parent_wave_id` column. Top-level wave discovery endpoints exclude subwaves, while `/waves/{id}/subwaves` lists child wave overviews. Subwave read access also requires the parent wave to be visible, and deleting a parent wave cascades through the API service to delete its subwaves.
 
+The waves v2 read boundary keeps timeline, reply-thread, and curation feeds as separate contracts. `/v2/waves/{id}/drops` returns the wave timeline feed, `/v2/drops/{id}/replies` returns the reply thread for a root drop after resolving its owning visible wave, and `/v2/waves/{id}/curations/{curation_id}/drops` returns drops for one wave curation.
+
 The waves v2 boundary also exposes `/v2/official-waves`, backed by the `official_waves` selector table. It returns readable `ApiWaveOverview` rows for listed wave ids and skips stale entries whose wave row no longer exists.
 
 Wave creators and wave admins can manage arbitrary wave metadata pairs through `/v2/waves/{id}/metadata`. Read access follows the same wave visibility rules as other wave v2 reads, while writes are restricted to the creator or members of the wave admin group. Metadata is stored in `waves_metadatas`, keyed by wave id and metadata key.
--- a/src/api-serverless/openapi.yaml
+++ b/src/api-serverless/openapi.yaml
@@ -2236,6 +2236,65 @@ paths:
                 $ref: "#/components/schemas/ApiDropAndWave"
         "404":
           description: Drop not found
+  /v2/drops/{id}/replies:
+    get:
+      tags:
+        - DropsV2
+      summary: Get V2 replies to a drop.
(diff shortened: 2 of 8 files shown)


punk6529:
@coderabbitai review

punk6529:
@coderabbitai review

punk6529:
@6529bot review

punk6529:
@6529bot review

punk6529:
@6529bot review

punk6529:
Follow-up on 6529bot's `limit=200` awareness note: I sanity-checked the downstream path and am leaving this as intentional/no-change.

- `findLatestDropsSimple` receives the validated amount and uses it as the bounded SQL limit.
- `findWaveFeed` maps exactly that bounded result list.
- `ApiDropMapper.mapDrops` batch-loads related data by the returned drop IDs; I did not find a hidden lower cap or fixed-size assumption that would make 200 unsafe.

So the behavioral change is real but intended: the documented max is 200, and `limit=200` now actually means 200 instead of silently collapsing to 50.

PR #1655, opened 2026-06-18
Fix profile CMS pointer migration execution

punk6529:
## Summary
- Fix the profile CMS pointer migration so db-migrate executes each CREATE TABLE statement separately.
- Make the new pointer/signature table creates idempotent so rerunning after the failed staging migration is safe.

## Validation
- node mock migration runner confirms two sequential runSql calls
- git diff --check --ignore-cr-at-eol
- npm run lint -- --max-warnings=0

Release train: profile-cms-phase-0-8

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration reliability. Migrations now handle existing tables gracefully without failure, making deployments and re-runs more robust.

### DIFF
--- a/migrations/20260617222400-profile-cms-pointer-events.js
+++ b/migrations/20260617222400-profile-cms-pointer-events.js
@@ -1,16 +1,30 @@
 const fs = require('fs');
 const path = require('path');
 
-exports.up = async function (db) {
-  const filePath = path.join(
-    __dirname,
-    'sqls',
-    '20260617222400-profile-cms-pointer-events-up.sql'
-  );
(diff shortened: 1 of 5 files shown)


punk6529:
/6529bot review

Focus request: migration execution safety after the failed staging dbMigrationsLoop run, idempotency if the first table was partially created, and whether splitting statements changes table definitions or rollback assumptions.

punk6529:
/6529bot review

Follow-up: replaced semicolon splitting with explicit one-statement SQL files, preserved idempotent up migrations, and added explicit down files. Please re-check migration execution safety and the failed-staging retry path.

PR #1672, opened 2026-06-25
[codex] Preserve wallet session cookies in API responses

prxt6529:
## What changed

- Normalize API Lambda HTTP responses so repeated `Set-Cookie` values are preserved in both `multiValueHeaders.Set-Cookie` and `cookies`.
- Remove `Set-Cookie` from the single-value `headers` map when repeated cookie values are present.
- Add handler-level regression tests for REST API Gateway and HTTP API cookie response shapes.

## Why

Multi-account wallet auth v2 relies on both the compatibility `6529_session` cookie and address-scoped `6529_session_<hash>` cookies surviving the deployed API response path. If production collapses repeated `Set-Cookie` headers to a single cookie, only the most recently signed wallet can refresh its web session. Switching back to another already-v2-authenticated wallet then incorrectly triggers the upgrade-auth prompt.

## Validation

- `npx jest src/api-serverless/src/handler.test.ts --runInBand --config /private/tmp/jest.no-container.config.js`
- `npx tsc --noEmit -p tsconfig.json`
- `npx eslint "src/**/*.ts" --max-warnings=0`

Note: the standard Jest command is blocked in this local environment because global setup requires a working container runtime for MySQL before even mocked handler tests run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP response handling to preserve `Set-Cookie` values across REST and HTTP API responses.
  * Normalized cookie output so duplicates are removed and cookie values consistently appear in the expected header/cookie fields for each API type.
  * Updated handling now supports cookie lists provided by the underlying HTTP layer.
* **Tests**
  * Expanded coverage for cookie propagation, normalization, and de-duplication behaviors across REST and HTTP API v2 events.

### DIFF
--- a/src/api-serverless/src/handler.test.ts
+++ b/src/api-serverless/src/handler.test.ts
@@ -191,7 +191,7 @@ describe('handler http responses', () => {
     expect(result.multiValueHeaders).toEqual({
       'Set-Cookie': [compatibilityCookie]
     });
-    expect(result.cookies).toEqual([compatibilityCookie]);
+    expect(result.cookies).toBeUndefined();
   });
 
   it('preserves repeated Set-Cookie headers for API Gateway REST responses', async () => {
@@ -230,10 +230,10 @@ describe('handler http responses', () => {
     expect(result.multiValueHeaders).toEqual({
       'Set-Cookie': [compatibilityCookie, scopedCookie]
     });
-    expect(result.cookies).toEqual([compatibilityCookie, scopedCookie]);
+    expect(result.cookies).toBeUndefined();
   });
 
-  it('copies normalized cookie arrays into multi-value headers', async () => {
+  it('copies normalized cookie arrays into REST multi-value headers', async () => {
     const compatibilityCookie = '6529_session=session-b.secret; Path=/api/auth';
     const scopedCookie = '6529_session_scoped=session-b.secret; Path=/api/auth';
     mockHttpHandler.mockResolvedValue({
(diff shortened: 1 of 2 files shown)


prxt6529:
Addressed the applicable review feedback in `25b34d0a`:

- changed `Set-Cookie` de-dupe to last-wins by cookie identity (`name` + `Path` + `Domain`), preserving same-name cookies on different paths/domains
- added pass-through coverage for no-cookie responses
- added single `Set-Cookie` header normalization coverage
- renamed the CodeRabbit-targeted test
- removed the virtual `serverless-http` mock

Validation run locally:
- `npx jest src/api-serverless/src/handler.test.ts --runInBand --config /private/tmp/jest.no-container.config.js`
- `npx tsc --noEmit -p tsconfig.json`
- `npx eslint "src/**/*.ts" --max-warnings=0`

prxt6529:
Hotfix follow-up in `972179a4` after staging exposed malformed REST API Gateway responses when `Set-Cookie` was present:

- REST API Gateway responses now emit repeated cookies only via `multiValueHeaders['Set-Cookie']`.
- HTTP API v2 responses emit repeated cookies via top-level `cookies`.
- Added/updated handler tests for both shapes.

Staging API redeployed from `1a-staging` merge `d5b0c06d`: https://github.com/6529-Collections/6529seize-backend/actions/runs/28160281311

Post-deploy validation against `api.staging.6529.io`:
- `POST /api/auth/session-login` with a throwaway generated wallet returned `201`.
- `Access-Control-Allow-Origin` was `https://staging.6529.io`.
- Response included two cookies: `6529_session` and address-scoped `6529_session_<hash>`.
- Access token response body was not logged.

prxt6529:
Fresh audit pass after the latest backend changes: no new actionable findings.

I re-read the final diff around `src/api-serverless/src/handler.ts` and `src/api-serverless/src/handler.test.ts` with the deployed REST behavior in mind. The response normalization now preserves repeated `Set-Cookie` values in the REST-compatible `multiValueHeaders['Set-Cookie']` shape, strips single-value `Set-Cookie` headers to avoid collapsed/duplicate emission, and keeps the HTTP API v2 `cookies` shape separate. Identity-based cookie dedupe is last-wins by cookie name/path/domain, which is the right behavior for duplicate session-cookie sources while preserving same-name cookies on different paths.

The remaining HTTP API v2 caveat is non-blocking for this PR: the current deployed auth path is REST, and the staging validation already exercised the REST path that this fix targets. I do not see a correctness, auth, CORS, or cookie-regression issue requiring another backend change.

PR #1673, opened 2026-06-26
Fix rememes S3 loop convergence

GelatoGenesis:
## Summary
- make `rememesS3Loop` converge by persisting existing legacy S3 assets, recording partial/unsupported media outcomes, and bounding transient retries
- add nullable rememe S3 processing status/error/attempt metadata and limit the S3 queue to unprocessed or retryable rows
- keep rememes loop functions on x86_64 so sharp-backed rememe jobs do not initialize on arm64
- return rememe compact media from scaled, thumbnail, original, then raw image fallback

## Testing
- `npm run lint`
- `npm test -- --runTestsByPath src/s3-rememes.test.ts`
- `npm run tsc -- -p tsconfig.json --noEmit`
- `npm run build`
- `cd src/api-serverless && npm run build`

## Deployment plan
1. Deploy and invoke `dbMigrationsLoop` first so the nullable rememe S3 processing columns exist.
2. Deploy `rememesLoop` next; this updates `rememesS3Loop`, `rememesRefreshMetadataLoop`, and `rememesFileLoop`.
3. Deploy `api` after that so the rememe media endpoint uses the S3 URL fallback query.

## Architecture docs
No `docs/architecture.md` update: this keeps the existing rememes loop/API surfaces and deployment units, with no new lambda, queue/topic, EventBridge trigger, or external integration.

## Summary by CodeRabbit

* **New Features**
  * Improved media handling for NFT/rememe images, with better fallback support for available image sizes and formats.
  * Added tracking for media processing status, retries, and errors to make asset handling more reliable.

* **Bug Fixes**
  * Reduced missed image uploads by reusing existing S3 assets more effectively.
  * Improved handling of unsupported or problematic media so items are marked clearly instead of failing silently.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,7 +52,7 @@
         "mysql": "^2.18.1",
         "node-cron": "^4.0.0",
         "node-emoji": "^2.2.0",
-        "node-fetch": "^2.7.0",
+        "node-fetch": "2.7.0",
         "openai": "^6.42.0",
         "passport": "^0.7.0",
         "passport-anonymous": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "mysql": "^2.18.1",
     "node-cron": "^4.0.0",
     "node-emoji": "^2.2.0",
-    "node-fetch": "^2.7.0",
+    "node-fetch": "2.7.0",
     "openai": "^6.42.0",
     "passport": "^0.7.0",
     "passport-anonymous": "^1.0.1",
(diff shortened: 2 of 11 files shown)


GelatoGenesis:
Operational follow-up for the 6529bot notes:

- Changed rememe fetch source intentionally clears all four `s3_image_*` URLs, status, error, last-attempt, and attempts in `rememeS3FieldsForRefresh`, which queues regeneration. Unchanged metadata image URLs preserve S3 state to avoid churn from volatile gateway URLs.
- `UNSUPPORTED` rows are terminal for the queue because `fetchMissingS3Rememes` only admits `NULL`, `TRANSIENT_ERROR`, or `PARTIAL` statuses.
- `node-fetch` is pinned at root to `2.7.0` (`npm ls node-fetch --depth=0`), so the current `{ timeout }` option and Node readable stream body handling are valid.
- The staging deploy order will be `dbMigrationsLoop` first, then `rememesLoop`, then `api`, so the nullable rememe S3 columns exist before the loop/API code that references them.
- The deploy workflow builds on `ubuntu-latest` and runs fresh `npm i` steps before `npm run build`, so the `dist/index.zip` artifact for `rememesLoop` is built on Linux x64 for the x86_64 Lambda runtime.

PR #1677, opened 2026-06-29
Add drop id to activity events

GelatoGenesis:
## Summary
- Adds nullable indexed `activity_events.drop_id` to support phase 2 indexed cleanup of drop-related activity events.
- Writes `drop_id` for newly recorded drop-created and drop-replied activity events.
- Leaves existing deletion behavior unchanged so old rows and current reads remain compatible until the backfill/switch PR.

## Deployment
- Deploy `dbMigrationsLoop` first so the nullable column is present and the migration-owned `activity_events_drop_id_idx` index is created with online DDL.
- Then deploy `tdhLoop`, `delegationsLoop`, `populateHistoricConsolidatedTdh`, and `api` so all drop-deletion/write paths share the updated entity shape.
- Phase 2 will backfill historic events and switch deletion to the indexed field.

## Validation
- `npm run lint`
- `npm run format`
- `npm run build`
- `npm run build` in `src/api-serverless`

### DIFF
--- a/migrations/20260629083000-add-activity-event-drop-id.js
+++ b/migrations/20260629083000-add-activity-event-drop-id.js
@@ -1,0 +1,42 @@
+'use strict';
+
+function ignoreMysqlError(promise, ignoredCodes) {
+  return promise.catch(function(error) {
+    if (error && ignoredCodes.indexOf(error.code) !== -1) {
+      return null;
+    }
+    throw error;
+  });
+}
+
+exports.up = function(db) {
+  return ignoreMysqlError(
+    db.runSql(
+      'ALTER TABLE activity_events ADD COLUMN drop_id varchar(100) DEFAULT NULL AFTER action'
+    ),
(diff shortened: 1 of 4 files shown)


GelatoGenesis:
Latest head `1acf8a82` addresses the stale 6529bot follow-up comments from `fe989333`:

- `src/entities/IActivityEvent.ts` now keeps `@Index('activity_events_drop_id_idx', { synchronize: false })`, so TypeORM does not own the index.
- `migrations/20260629083000-add-activity-event-drop-id.js` now creates the index with `ALTER TABLE activity_events ADD INDEX activity_events_drop_id_idx (drop_id), ALGORITHM=INPLACE, LOCK=NONE`.
- The rollback still ignores `ER_CANT_DROP_FIELD_OR_KEY`, matching MySQL for a missing index/column.

All current PR checks are green on `1acf8a82`.

PR #1679, opened 2026-06-29
Backfill activity event drop ids

GelatoGenesis:
## Summary
- Backfill historic `activity_events.drop_id` values for `DROP_CREATED` and `DROP_REPLIED` events from their JSON payloads.
- Run the historic backfill in bounded primary-key batches so each update commits separately instead of holding locks for the entire migration.
- Replace drop-feed activity deletion for drops with indexed predicates on `drop_id` and `target_id`, removing the `activity_events.data LIKE` scan.

## Deployment
- Deploy `dbMigrationsLoop` first so historic rows are backfilled before the API starts relying on `drop_id` for deletion.
- Then deploy `tdhLoop`, `delegationsLoop`, `populateHistoricConsolidatedTdh`, and API from the same branch/environment.

## Validation
- `npm run format`
- `npm run lint`
- `npm test -- src/drops/drops-db.test.ts --runInBand`
- `npm run build`
- `cd src/api-serverless && npm run build`

### DIFF
--- a/migrations/20260629122000-backfill-activity-event-drop-id.js
+++ b/migrations/20260629122000-backfill-activity-event-drop-id.js
@@ -1,0 +1,98 @@
+'use strict';
+
+var BATCH_SIZE = 5000;
+var BATCH_SIZE_BIGINT = BigInt(BATCH_SIZE);
+
+function getId(value) {
+  if (value === null || value === undefined) {
+    return null;
+  }
+  return BigInt(value);
+}
+
+async function backfillAction(db, action, dataPath) {
+  var bounds = await db.runSql(
+    'select min(id) as min_id, max(id) as max_id from activity_events where drop_id is null and action = ?',
+    [action]
+  );
+  var firstRow = bounds && bounds[0] ? bounds[0] : {};
+  var minId = getId(firstRow.min_id);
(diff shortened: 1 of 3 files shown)


GelatoGenesis:
Current head `d1be26d8` addresses the earlier 6529bot findings and I verified the remaining dataflow concern against the activity-event writer code before merge.\n\n- `ActivityEventAction` currently has only `DROP_CREATED`, `DROP_REPLIED`, and `WAVE_CREATED`.\n- `ActivityRecorder.recordDropCreated` writes created-drop events with the created drop id in `data.drop_id`; the backfill copies that for `DROP_CREATED`.\n- For replies, `ActivityRecorder.recordDropCreated` writes `target_type = DROP` / `target_id = reply_to.drop_id` for the parent drop and `data.reply_id = drop_id` for the reply itself; the new delete predicate covers parent cleanup by target columns and reply cleanup by the backfilled `drop_id`.\n- `WAVE_CREATED` writes only `data.wave_id` and has no drop id to backfill/delete.\n- The migration now documents the intentionally irreversible `down` and the committed partial-progress/resume behavior; it also logs batch progress.\n\nAll PR checks are green after the latest rebase: Build backend and API, DCO, SonarCloud, Snyk, and CodeRabbit.

PR #1686, opened 2026-07-01
Add drop voting summary indexes

GelatoGenesis:
## Summary
- add online db-migrate indexes for drop voting summary lookups and rank ordering inputs
- mark the new composite indexes as migration-managed in TypeORM metadata
- add a TypeORM declaration merge so composite indexes can use the already-supported synchronize:false option

## Risk
- index-only change; no API query semantics, vote logic, or winner conversion logic changed

## Verification
- npm run lint
- npm run build

## Docs
- docs/architecture.md not updated; this does not add, remove, rename, or rewire any Lambda, API boundary, SQS/SNS resource, DB runtime pattern, deployable service, or external integration

### DIFF
--- a/migrations/20260701093000-add-drop-voting-summary-indexes.js
+++ b/migrations/20260701093000-add-drop-voting-summary-indexes.js
@@ -13,6 +13,10 @@ function addIndex(db, sql) {
   return ignoreMysqlError(db.runSql(sql), ['ER_DUP_KEYNAME']);
 }
 
+function dropIndex(db, sql) {
+  return ignoreMysqlError(db.runSql(sql), ['ER_CANT_DROP_FIELD_OR_KEY']);
+}
+
 exports.up = function(db) {
   return addIndex(
     db,
@@ -44,7 +48,36 @@ exports.up = function(db) {
     });
 };
 
-exports.down = function() {};
+exports.down = function(db) {
+  return dropIndex(
+    db,
+    'DROP INDEX idx_drops_wave_type_created_id ON drops'
+  )
+    .then(function() {
+      return dropIndex(
+        db,
+        'DROP INDEX idx_wave_leaderboard_entries_wave_vote_time_drop ON wave_leaderboard_entries'
+      );
+    })
+    .then(function() {
+      return dropIndex(
+        db,
+        'DROP INDEX idx_drop_ranks_wave_vote_last_drop ON drop_ranks'
+      );
+    })
+    .then(function() {
+      return dropIndex(
+        db,
+        'DROP INDEX idx_winner_drop_voter_votes_drop_votes_voter ON winner_drop_voter_votes'
+      );
+    })
+    .then(function() {
+      return dropIndex(
+        db,
+        'DROP INDEX idx_drop_voter_states_drop_votes_voter ON drop_voter_states'
+      );
+    });
+};
 
 exports._meta = {
   version: 1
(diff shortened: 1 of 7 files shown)


GelatoGenesis:
Addressed the 6529bot findings in b427f1f4:

- Added a real reversible down migration that drops the five indexes and ignores ER_CANT_DROP_FIELD_OR_KEY.
- Broadened the TypeORM declaration merge from synchronize?: false to synchronize?: boolean.
- Verified deployed DB engine via AWS RDS: us-east-1 production cluster seize-prod-nova is aurora-mysql 8.0.mysql_aurora.3.10.3. These are secondary indexes on existing InnoDB/Aurora MySQL tables, and the migration intentionally keeps ALGORITHM=INPLACE, LOCK=NONE so deploy fails rather than silently falling back to a locking table copy if online DDL is not supported. eu-west-1 has no app DB cluster listed.

GelatoGenesis:
Addressed the remaining 6529bot migration concern in 29d06943:

-  now uses the same migration-managed index list as .
- On any non-duplicate ADD INDEX failure,  runs reverse-order cleanup for all five managed indexes and then rethrows the original error.
-  reuses that same reverse-order cleanup path.

Re-verified locally:
- npm run lint
- npm run build

GelatoGenesis:
Correction to the previous comment: addressed the remaining 6529bot migration concern in 29d06943.

- `exports.up` now uses the same migration-managed index list as `exports.down`.
- On any non-duplicate `ADD INDEX` failure, `exports.up` runs reverse-order cleanup for all five managed indexes and then rethrows the original error.
- `exports.down` reuses that same reverse-order cleanup path.

Re-verified locally:
- `npm run lint`
- `npm run build`

GelatoGenesis:
Addressed the `drop_ranks.vote` review concern in ce92ce78 by removing the proposed `drop_ranks` composite index from this index-only phase.

Rationale:
- `drop_ranks.vote` is stored as `varchar(100)`, while the ranking semantics are numeric.
- Existing realtime leaderboard/ranking SQL either needs numeric ordering explicitly or joins `drop_ranks` by `drop_id`, so `(wave_id, vote, last_increased, drop_id)` is not a clean index-only optimization.
- Keeping it out avoids adding an index that could be unused or misleading until a separate schema/query change handles `drop_ranks.vote` properly.

Remaining migration indexes are the lower-risk ones for the voting summary paths:
- `drop_voter_states (drop_id, votes, voter_id)`
- `winner_drop_voter_votes (drop_id, votes, voter_id)`
- `wave_leaderboard_entries (wave_id, vote, timestamp, drop_id)`
- `drops (wave_id, drop_type, created_at, id)`

Re-verified locally after this change:
- `npm run lint`
- `npm run build`

GelatoGenesis:
Addressed the 6529bot advisory cleanup edge cases in ccc4fb08.

Changes:
- `addIndex` now returns whether the current `up` run actually created the index, rather than treating duplicate index names the same as newly-created indexes.
- `exports.up` tracks only indexes created during that specific run and only cleans those up on a later non-duplicate failure.
- If cleanup itself fails, the original add-index error is preserved and the cleanup error is attached as `cleanupError` instead of masking the root cause.
- `exports.down` still drops all migration-managed index names intentionally, for an explicit rollback of this migration.

Re-verified locally:
- `npm run lint`
- `npm run build`

PR #1694, opened 2026-07-03
Optimize wave overview candidate lookups

GelatoGenesis:
## Summary
- Add a short-lived global candidate cache for wave overview ranking queries.
- Keep per-user pin, mute, and followed filtering live after candidate selection.
- Re-read selected wave ids through the existing DB visibility filter before returning results, falling back to the old SQL path if the candidate window is not sufficient.
- Add wave metric score-order indexes for scored overview candidate scans and a reader/wave index for live mute probes.

## Safety
- Candidate cache keys include request-visible global filters and sorted eligible group ids.
- Cached candidates are ranking hints only; final returned waves still pass current DB visibility checks.
- Followed-only requests, deep pages, authenticated scored filter requests, incomplete candidate windows, and full windows containing muted candidates continue using the existing SQL path.
- Candidate tie-breaks now use deterministic id-desc ordering matching the SQL fallback instead of locale-aware JS ordering.

## Schema deployment
- This repository applies entity-index changes through `dbMigrationsLoop` with TypeORM `syncEntities: true`; no manual migration file is used for this repo path unless explicitly requested.
- Deploy `dbMigrationsLoop` before API so the new indexes are applied before the API uses the candidate query path.

## Validation
- `npm test -- --runTestsByPath src/api-serverless/src/waves/wave-overview-candidate-cache.test.ts src/api-serverless/src/waves/waves-api-db.test.ts`
- `npm run lint`
- `npm run build`

### DIFF
--- a/src/api-serverless/src/waves/wave-overview-candidate-cache.test.ts
+++ b/src/api-serverless/src/waves/wave-overview-candidate-cache.test.ts
@@ -1,0 +1,81 @@
+import { redisGet, redisSetJson } from '@/redis';
+import { withWaveOverviewCandidateCache } from './wave-overview-candidate-cache';
+
+jest.mock('@/redis', () => ({
+  redisGet: jest.fn(),
+  redisSetJson: jest.fn()
+}));
+
+const redisGetMock = jest.mocked(redisGet);
+const redisSetJsonMock = jest.mocked(redisSetJson);
+
+describe('wave overview candidate cache', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('returns cached candidates without calling the loader', async () => {
+    const cached = [
+      {
+        waveId: 'wave-1',
+        tierRank: 1,
+        sortVal: 100,
+        latestDropTimestamp: 10
+      }
+    ];
+    redisGetMock.mockResolvedValue(cached);
+    const getValue = jest.fn();
+
+    await expect(
+      withWaveOverviewCandidateCache({
+        keyParts: {
+          overviewType: 'scored',
+          eligibleGroups: ['group-1']
+        },
+        getValue
+      })
+    ).resolves.toBe(cached);
+
+    expect(getValue).not.toHaveBeenCalled();
+    expect(redisSetJsonMock).not.toHaveBeenCalled();
+  });
+
+  it('coalesces equivalent misses and writes the loaded candidates', async () => {
+    redisGetMock.mockResolvedValue(null);
+    const loaded = [
+      {
+        waveId: 'wave-1',
+        tierRank: 1,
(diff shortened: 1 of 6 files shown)


GelatoGenesis:
Addressed the blocking bot findings in the latest head:

- Muted candidate windows: the candidate path now returns null and falls back to legacy SQL whenever a full candidate window contains any muted wave. Added tests for recently-dropped and scored variants.
- Ordering drift: replaced localeCompare tie-breaks with deterministic id-desc comparison matching the SQL fallback. Added a tie-order test.
- Mute probe index: added idx_wrm_reader_wave on (reader_id, wave_id); pinned waves already use primary key order (profile_id, wave_id).
- Index rollout: per repo AGENTS, schema/index changes are applied by dbMigrationsLoop through TypeORM syncEntities=true. PR body now calls out dbMigrationsLoop before API for staging/prod.

GelatoGenesis:
Addressed the current follow-up findings in c2a0a117:\n\n- Removed the page materialization type predicate that asserted WaveEntityWithLastDropTime. The candidate path now builds an explicit WaveEntity[] accumulator after the final visibility read.\n- Added a count-only info log for visibility-mismatch fallbacks: [WAVE_OVERVIEW_CANDIDATE_FALLBACK] with reason, candidate_count, visible_count, and eligible_group_count. This keeps the safe legacy fallback observable in CloudWatch without logging wave ids or group ids.\n- Confirmed eligible groups are sorted into the candidate cache key, so different request group sets do not share the same candidate key. The final visibility read remains the correctness gate for stale/deleted/visibility-changed rows.\n- Confirmed index rollout is via dbMigrationsLoop with TypeORM syncEntities=true in src/dbMigrationsLoop/index.ts, per repo AGENTS.\n- Confirmed PINNED_WAVES_TABLE is covered by the PinnedWaveEntity primary key order (profile_id, wave_id).\n\nLocal validation after this change:\n- npm test -- --runTestsByPath src/api-serverless/src/waves/wave-overview-candidate-cache.test.ts src/api-serverless/src/waves/waves-api-db.test.ts\n- npm run lint\n- npm run build

GelatoGenesis:
Addressed the muted partial-window follow-up in 3f7255b4:\n\n- Added seeded comparisons for a muted high-score wave in a sub-candidateLimit window for both recently-dropped and scored overview paths.\n- Each test asserts the candidate path returns the same ordering as a forced legacy SQL path, and that the muted wave is demoted behind the visible lower-score wave.\n\nLocal validation after this follow-up:\n- npm test -- --runTestsByPath src/api-serverless/src/waves/wave-overview-candidate-cache.test.ts src/api-serverless/src/waves/waves-api-db.test.ts\n- npm run lint\n- npm run build

PR #1511, opened 2026-04-15
pnpm - SFW - 6529 shim

prxt6529:
## Summary by CodeRabbit

* **New Features**
  * Repo-local CLI wrapper and repo-controlled pnpm workflow; direct npm/npx/yarn use is blocked.
  * Unified install/build/deploy commands across services.

* **Chores**
  * CI/CD updated to Node 22 and Corepack/pnpm with caching; many packages now declare pnpm as the package manager.
  * Workspace/package management and deployment packaging streamlined.

* **Docs**
  * Developer guides and README updated for bootstrap and new workflow.

* **Tests**
  * Added/updated unit tests and improved deploy UI status handling (startup-failure badges).

PR #1698, opened 2026-07-04
Enforce npm as the pinned package manager

punk6529:
## Issue
- PR #1511 attempted an npm→pnpm migration with a bespoke CLI shim layer; it went stale (640 commits behind, 16 conflicts, 7 newer services never converted) and was closed.
- The enforcement goal it chased is still valid: one package manager, consistently pinned, with no way for a new service or a stray `pnpm-lock.yaml`/`yarn.lock` to drift in unnoticed.

## Fix
- Keep npm (no migration) and enforce discipline with a single dependency-free checker script plus a fast-fail CI step — roughly 90% of the enforcement value of #1511 at a fraction of the surface.

## Changes
- New `scripts/check-package-manager.mjs`: validates every tracked `package.json` (root + all 55 service/module manifests) carries `"packageManager": "npm@10.9.8"`, and that no forbidden package-manager files are tracked (`yarn.lock`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `bun.lockb`, `.yarnrc*`). `--fix` mode applies the pin everywhere, so onboarding a new service is one command and CI catches omissions by construction.
- `.github/workflows/on-pull-request.yml`: new "Check package manager discipline" step before dependency install (script has zero dependencies, so it fails fast).
- Applied the pin to all 56 `package.json` files.

## Validation
- `node scripts/check-package-manager.mjs` passes on the branch (56 files pinned, no stray lockfiles).
- Negative tests: removing a pin and adding a tracked `pnpm-lock.yaml` each fail the check with actionable messages; `--fix` restores the pin.
- `npx eslint scripts/check-package-manager.mjs --max-warnings=0` clean; actionlint clean on the workflow; YAML parses.
- Not tested: corepack behavior on developer machines (the pin is advisory for plain npm and only binding under corepack — that is the intended 90/10 trade-off).

## Risk
- Level: Low
- Why: no runtime, dependency, or deploy-config changes; the pin field is inert metadata for npm itself; CI gains one fast read-only step.
- Rollback: revert the commit.

## Deployment
- None. CI-only change; no lambdas or services need redeploying.

## Review Notes
- npm version pinned to 10.9.8 to match the npm bundled with the Node 22 line used by PR CI (`actions/setup-node` `node-version: 22`), not the newest npm 11 — happy to bump if the team prefers tracking npm 11.
- `packageManager` is intentionally the only enforcement (no `engines`/`engine-strict`, no shim wrappers) to keep the surface minimal per the #1511 closure note.

## Summary by CodeRabbit

* **New Features**
  * Added a repo-wide npm version pin to keep package management consistent.
  * Included a CI check that verifies the expected package manager version during pull requests.

* **Bug Fixes**
  * Updated many project manifests to declare the same npm version, reducing setup and build drift.
  * Added support for automatically correcting mismatched package manager settings when needed.

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -18,6 +18,8 @@ jobs:
         with:
           node-version: '22'
           cache: 'npm'
+      - name: Check package manager discipline
+        run: node scripts/check-package-manager.mjs
       - name: Install dependencies
         run: npm i && pushd src/api-serverless && npm i && popd
       - name: Verify generated files are committed
--- a/package.json
+++ b/package.json
@@ -151,5 +151,6 @@
     "minimatch@5": {
       "brace-expansion": "^2.0.2"
     }
-  }
+  },
+  "packageManager": "npm@10.9.8"
 }
--- a/scripts/check-package-manager.mjs
+++ b/scripts/check-package-manager.mjs
@@ -1,0 +1,89 @@
+import { execSync } from 'node:child_process';
+import console from 'node:console';
+import fs from 'node:fs';
+import process from 'node:process';
+
+const PINNED_PACKAGE_MANAGER = 'npm@10.9.8';
+const FORBIDDEN_FILES = [
+  'yarn.lock',
+  'pnpm-lock.yaml',
+  'pnpm-workspace.yaml',
+  'bun.lockb',
+  'bun.lock',
+  '.yarnrc',
+  '.yarnrc.yml'
+];
+
+const fix = process.argv.includes('--fix');
+
+const trackedFiles = execSync('git ls-files', { encoding: 'utf8' })
+  .split('\n')
+  .filter(Boolean);
+
+const packageJsonFiles = trackedFiles.filter(
+  (file) => file === 'package.json' || file.endsWith('/package.json')
+);
+
+const errors = [];
+
+const strayLockfiles = trackedFiles.filter((file) => {
+  const basename = file.split('/').pop();
+  return FORBIDDEN_FILES.includes(basename);
+});
+for (const file of strayLockfiles) {
+  errors.push(
+    `${file}: forbidden package-manager file. This repo uses npm only; remove it.`
+  );
+}
+
+for (const file of packageJsonFiles) {
+  const raw = fs.readFileSync(file, 'utf8');
+  let manifest;
+  try {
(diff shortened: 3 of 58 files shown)


punk6529:
Addressed the review findings in ea9184d9:

- **--fix reserialization (Important)**: fixed — `--fix` now surgically inserts or regex-replaces only the `packageManager` key, preserving the rest of the manifest byte-for-byte (verified: removing the pin and re-running `--fix` produces a zero-line diff against the original file, including EOL style and trailing newline).
- **`git ls-files` without cwd / raw errors**: fixed by removing `child_process` entirely — the script now walks the tree from the repo root derived from its own location, so it is CWD-independent and has no PATH-resolved subprocess (this also resolves the SonarCloud S4036 vulnerability that failed the quality gate). Side benefit: untracked stray lockfiles are now caught too, not just tracked ones.
- **Basename matching for forbidden files (deferred)**: keeping basename matching deliberately — a `bun.lock` or `yarn.lock` anywhere in the tree is a real signal regardless of directory, `.yarnrc.example` does not match (`.yarnrc` is matched exactly), and build/dependency directories are excluded from the walk. Will anchor to specific paths if a false positive ever shows up.
- **Corepack enforcement (deferred)**: intentional 90/10 trade-off per the PR body and the #1511 closure note — the pin is binding for corepack-enabled dev machines and documents intent everywhere else, without adding a network-dependent corepack download step to CI. Happy to add `corepack enable` in a follow-up if the team wants hard enforcement.

punk6529:
Follow-up findings addressed in 6032d5ef:

- **Gitignored files scanned (Important)**: fixed — the walk now honors the root `.gitignore` (dependency-free matcher covering the pattern shapes the file uses: bare names, `dir/` suffixes, leading-`/` anchors, `*`/`?`/`**` globs). Local scratch directories (e.g. `/temp`, `*.ignore`) are neither validated nor rewritten; untracked but non-ignored files are still caught. Negation patterns are deliberately not applied — that can only make the check skip more, never fail a re-included file (the repo's only negation is `!.env.sample`).
- **`lastIndexOf('}')` insert anchor (deferred)**: safe by construction — the insert path only runs after `JSON.parse(raw)` succeeds, and for any valid top-level JSON object the last `}` in the text *is* the closing brace (nothing but whitespace may follow it). Unparseable manifests error out before reaching the writer.

PR #1701, opened 2026-07-04
perf(ownersBalancesLoop): group owners once and batch consolidation key lookups

punk6529:
## Issue

`ownersBalancesLoop` does two kinds of avoidable work on every run, and both explode on reset runs:

1. `updateOwnerBalances` filters the **entire** `owners` array once **per address** (`owners.filter(o => equalIgnoreCase(o.wallet, address))`). On a reset that is (all addresses) × (all owner rows) case-insensitive comparisons.
2. `consolidateOwnerBalances` queries the `address_consolidation_keys` view **once per address** inside an unbounded `Promise.all` — one DB round trip per wallet, even though the query helper already accepts an address array. The reset path additionally hydrates every `owner_balances` entity just to read `.wallet`.

Part of a batch of performance-only PRs from a full backend audit; all changes are intended to be **output-identical**.

## Fix

- **Group once, look up per address**: build `Map<lowercased wallet, NFTOwner[]>` in one pass over `owners`, then `get(address.toLowerCase())` per address. Equivalent because: the old `equalIgnoreCase` never matched falsy wallets (falsy-walleted rows are skipped when building the map); per-wallet row order is the `owners` array order, same as `filter` preserved; a missing wallet yields `[]`, feeding the same `deleteDelta` branch as an empty filter result.
- **Batch the consolidation-key view lookups**: fetch keys for all addresses in `5000`-address chunks (`fetchWalletConsolidationKeysViewForWallet` already takes `string[]`), build `Map<address, consolidation_key>`, then resolve per address in memory. Fallback semantics preserved exactly: address not in the view → `consolidationKey = address.toLowerCase()`, `consolidationAddresses = [address.toLowerCase()]`, same as the old `[0]`-undefined branch. First row wins per address, matching the old per-address `[0]`. (View addresses are stored lowercased — see the 20240619220628 migration — and every address added to the `addresses` set is lowercased at insertion, so map key casing is exact.)
- **Reset path fetches only wallets**: new `fetchAllOwnerBalancesWallets()` selects the `wallet` column (`getRawMany`) instead of `getMany()` on full entities. Same strings end up in the same `Set`.

## Changes

- `src/ownersBalancesLoop/owners_balances.ts` — owner grouping map; chunked view lookup + in-memory key resolution; reset path uses the wallet-only query.
- `src/ownersBalancesLoop/db.owners_balances.ts` — adds `fetchAllOwnerBalancesWallets()`. `fetchAllOwnerBalances` itself is left untouched (still used by `tdh_upload.ts`).
- `src/ownersBalancesLoop/owners-balances.test.ts` — **new**: aggregation equivalence with mixed-case wallets (the exact thing `equalIgnoreCase` used to cover), delete-delta behavior, batched-view fallback semantics, and 5000-chunking.

## Validation

- `npx jest src/ownersBalancesLoop/owners-balances.test.ts` — 3/3 pass.
- `npx tsc --noEmit` clean for touched files (pre-existing unrelated `pdf-lib`/`zod` errors in untouched dirs); `eslint --max-warnings=0` + `prettier --check` clean.
- Not tested live: a reset-mode loop run against production-size data.

## Risk

- Level: **Low**
- Why: batch job only (no API surface); rewrites are mechanical with fallback semantics preserved and tested; chunk size 5000 is well under typical `max_allowed_packet` for a list of 42-char strings.

## Deployment

Not to be deployed yet — awaiting human review (@gelatogenesis). When approved: `ownersBalancesLoop` only.

## Review Notes

- One pre-existing oddity spotted while auditing (NOT changed here, flagging for a decision): `fetchAllNftOwners(contracts?, pk?)` in `src/nftOwnersLoop/db.nft_owners.ts` calls `.where()` twice, so when both args are passed (as `updateOwnerBalances` does) TypeORM silently **drops the contract filter**. Harmless today if `nft_owners` only contains tracked contracts, but `.andWhere` was almost certainly intended. Fixing it changes fetched rows, so it's out of scope for this behavior-identical PR.
- Update (review round 1): consolidations are now deduped by key before the fan-out, so `getConsolidatedBalances` runs once per unique consolidation instead of once per member address, and `WalletConsolidationKey` was corrected to the view's real column shape (`address`), removing a cast. The remaining per-consolidation `fetchAllNftOwners` call inside `getConsolidatedBalances` is still one query per consolidation — left alone deliberately: batching it would hold every owner row of every consolidation in memory at once on reset runs, which is a trade-off rather than a free win.

## Summary by CodeRabbit

* **New Features**
  * Improved owner balance updates for large wallet sets, including more efficient handling during consolidation.

* **Bug Fixes**
  * Fixed wallet matching to be case-insensitive, ensuring balances group correctly.
  * Strengthened consolidation-key resolution with chunked lookups and safer fallback behavior when consolidation data is missing.

* **Tests**
  * Added Jest coverage for incremental updates and consolidation chunking, including edge cases and fallback scenarios.

### DIFF
--- a/src/entities/IDelegation.ts
+++ b/src/entities/IDelegation.ts
@@ -89,6 +89,7 @@ export interface DelegationEvent extends ConsolidationEvent {
 export class NFTDelegationBlock extends BlockEntity {}
 
 export interface WalletConsolidationKey {
-  wallet: string;
+  // matches the address_consolidation_keys view columns
+  address: string;
   consolidation_key: string;
 }
--- a/src/ownersBalancesLoop/db.owners_balances.ts
+++ b/src/ownersBalancesLoop/db.owners_balances.ts
@@ -32,6 +32,15 @@ export async function fetchAllOwnerBalances(addresses?: string[]) {
   return await queryBuilder.getMany();
 }
 
+export async function fetchAllOwnerBalancesWallets(): Promise<string[]> {
+  const rows: { wallet: string }[] = await getDataSource()
+    .getRepository(OwnerBalances)
+    .createQueryBuilder('ownerBalances')
+    .select('ownerBalances.wallet', 'wallet')
+    .getRawMany();
+  return rows.map((r) => r.wallet);
+}
+
 export async function fetchAllOwnerBalancesMemes(addresses?: string[]) {
   const queryBuilder = getDataSource()
     .getRepository(OwnerBalancesMemes)
--- a/src/ownersBalancesLoop/owners-balances.test.ts
+++ b/src/ownersBalancesLoop/owners-balances.test.ts
@@ -1,0 +1,235 @@
+import {
+  fetchAllSeasons,
+  fetchTransactionAddressesFromBlock,
+  fetchWalletConsolidationKeysViewForWallet
+} from '../db';
+import { NFTOwner } from '../entities/INFTOwner';
+import { MemesSeason } from '../entities/ISeason';
+import {
+  fetchAllNftOwners,
+  getMaxNftOwnersBlockReference
+} from '../nftOwnersLoop/db.nft_owners';
+import {
+  fetchAllOwnerBalancesWallets,
+  fetchRefreshOutdatedBalances,
+  getMaxOwnerBalancesBlockReference,
+  persistConsolidatedOwnerBalances,
+  persistOwnerBalances
+} from './db.owners_balances';
+import {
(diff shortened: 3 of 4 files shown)


punk6529:
@coderabbitai review

punk6529:
@coderabbitai review

punk6529:
@6529bot security

GelatoGenesis (approved):

PR #1702, opened 2026-07-04
Polish Meme edition size floor plumbing

punk6529:
## Issue
- Review follow-ups from #1692 (all noted in its review thread as polish-level):
  1. The Manifold lazy-claim contract address and ABI were duplicated between `src/memes-edition-size-floor.ts` and `src/mintAnnouncementsLoop/manifold-claim.service.ts`, so a change could drift.
  2. `MEMES_EDITION_SIZE_FLOOR_CAP = 310` lived in the resolver module with its rationale only in the PR body.
  3. Floors were refreshed only for the single latest Meme: if the nftsLoop were down across an entire drop window, that drop's floor would stay at its migration-backfilled supply forever once a newer Meme existed.

## Fix
- Extract the shared contract constants, relocate the cap next to the other Memes business constants with an explanatory comment, and widen the refresh set to "latest Meme + any Meme minted in the last 30 days" so missed refreshes self-heal.

## Changes
- New `src/constants/manifold.ts` with `MANIFOLD_LAZY_CLAIM_CONTRACT` and `MANIFOLD_LAZY_CLAIM_ABI`; both consumers now import it (re-exported via `@/constants`).
- `MEMES_EDITION_SIZE_FLOOR_CAP` moved to `src/constants/index.ts` with the 310-editions rationale.
- `getMemeTokenIdsForEditionSizeFloorRefresh` now takes `mint_date` into account: refreshes the latest Meme (unconditionally, even with a missing mint_date) plus any Meme minted within a 30-day window; injectable `nowMillis` keeps it pure and testable.
- Tests extended: outage-recovery case (non-latest Meme inside the window is refreshed), outside-window case, and missing-mint_date case.

## Validation
- `npx jest src/nftsLoop/nfts.test.ts src/memes-edition-size-floor.test.ts src/nftsLoop/nft-extended-data.test.ts`: 26/26 pass.
- eslint clean on all changed files.
- No schema, API-contract, or generated-file changes.

## Risk
- Level: Low
- Why: behavior-preserving refactors plus a strictly-wider refresh set; the widened window adds at most a handful of extra bounded, timeboxed RPC lookups per loop run (concurrency 10, 10s timeout, failure falls back to existing floor).
- Rollback: revert the commit; floors persist in the DB and are unaffected.

## Deployment
- `nftsLoop` (refresh window + shared constants), `mintAnnouncementsLoop` (shared constants import). No migration. `tdhLoop`/`api` unaffected at runtime (pure import-path changes compile identically).

## Review Notes
- The 30-day window intentionally covers multiple missed drops; old fully-minted Memes stay excluded, preserving #1692's forward-looking design.

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -2,9 +2,16 @@ import { Network } from '@/alchemy-sdk';
 
 export * from './db-tables';
 export * from './delegations';
+export * from './manifold';
 export * from './media';
 
 export const MEMES_CONTRACT = '0x33FD426905F149f8376e227d0C9D3340AaD17aF1';
+
+// Upper bound applied to Manifold claim totalMax when deriving a Meme's
+// edition-size floor. Meme drops are sized at no more than 310 editions, so
+// any larger on-chain totalMax is a placeholder/unbounded claim value rather
+// than a real edition size (see nfts.edition_size_floor).
+export const MEMES_EDITION_SIZE_FLOOR_CAP = 310;
 export const GRADIENT_CONTRACT = '0x0c58ef43ff3032005e472cb5709f8908acb00205';
 export const MEMELAB_CONTRACT = '0x4db52a61dc491e15a2f78f5ac001c14ffe3568cb';
 export const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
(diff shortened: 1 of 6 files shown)


punk6529:
C:/Program Files/Git/6529bot review general

punk6529:
@6529bot general

punk6529:
@6529bot review general

PR #1703, opened 2026-07-04
perf(subscriptions): batch eligibility lookups and replace per-subscription table scans

punk6529:
## Issue

`fetchSubscriptionEligibility(key)` costs two queries per call — one of them (`SELECT MAX(id) FROM memes_seasons`) identical every time. Two hot paths called it **once per subscription inside `Promise.all`**:

- the public **upcoming meme subscription counts** API endpoint (via `fetchEffectiveUpcomingMemeSubscriptionCounts`, which loads *every* auto-subscription) — thousands of queries per HTTP request against the API's 10-connection read pool;
- the **daily subscriptions job** (`populateAutoSubscriptionsForMemeId` and `createFinalSubscriptions`), which additionally scanned the entire `subscription_balances` result linearly per subscription (`balances.find(equalIgnoreCase(...))` — O(subscriptions × balances)) and fetched the airdrop address for *every* subscription even when the result was discarded.

Part of a batch of performance-only PRs from a full backend audit; all changes intended to be **output-identical**.

## Fix

- **New `fetchSubscriptionEligibilityForKeys(keys): Map<lowercased key, count>`** — one `MAX(id)` query + one `IN (...)` query per 5000 distinct keys. Semantics identical to the single-key version per key: no seasons → 1; key absent or `sets` falsy/0 → 1; otherwise `sets`. The single-key `fetchSubscriptionEligibility` now delegates to it — same two queries, same results, so the remaining per-identity call sites are untouched behaviorally.
- **API (`api.subscriptions.db.ts`)**: the per-autoSub `Promise.all` becomes one batch call. The map it produces is keyed by lowercased consolidation key with a default of 1 for every requested key — exactly the shape the downstream `getAutoSubscriptionEffectiveCount` (`.get(key.toLowerCase()) ?? 1`) already expects.
- **Daily job (`subscriptionsDaily/subscriptions.ts`)**:
  - balances and auto-subscriptions are now first-match-preserving lowercased maps (`.find` with `equalIgnoreCase` was first-match; keys are unique in both tables, and first-wins insertion preserves the tie-break exactly if they ever weren't);
  - eligibility is batched once for all filtered subscriptions;
  - the `fetchAirdropAddressForConsolidationKey` call moved **inside** the `balance >= mint price` branch — its result was only read there, so no-balance/insufficient-balance subscriptions no longer pay a TDH-wallet + delegations query (read-only calls; no side effects);
  - the auto-subscription delta check uses a lowercased `Set` instead of a nested `some(equalIgnoreCase(...))` scan (falsy-key handling preserved: falsy keys never matched before and are still always kept in the delta).

## Changes

- `src/subscriptionsDaily/db.subscriptions.ts` — adds `fetchSubscriptionEligibilityForKeys`; `fetchSubscriptionEligibility` delegates to it.
- `src/subscriptionsDaily/subscriptions.ts` — maps + batched eligibility + airdrop lookup moved into the funded branch.
- `src/api-serverless/src/subscriptions/api.subscriptions.db.ts` — batched eligibility for auto subscriptions.
- `src/subscriptionsDaily/fetch-subscription-eligibility-for-keys.test.ts` — **new** unit tests: defaulting, lowercasing, dedup, 5000-chunking, no-seasons behavior, and single-key wrapper equivalence.

## Validation

- New tests: 6/6 pass.
- Existing `src/api-serverless/src/subscriptions/api-subscriptions-db.test.ts` (covers `fetchUpcomingMemeSubscriptionCounts`, the changed API path) and `src/subscriptionsDaily/subscriptions.test.ts`: 13/13 pass unchanged.
- Not tested live: a full daily-subscriptions run against production data.

## Risk

- Level: **Low–Medium** (touches a money-adjacent flow — final subscription lists — hence the extra test coverage and the strict semantics-preservation notes above)
- Why medium-ish: `createFinalSubscriptions` decides mint allocations; though every rewrite is mechanical (find→map, per-item→batch), reviewer scrutiny should focus here.

## Deployment

Not to be deployed yet — awaiting human review (@gelatogenesis). When approved: `api`, `subscriptionsDaily`, and `transactionsProcessingLoop` (imports the shared `db.subscriptions` module, even though its own call sites are unchanged).

## Review Notes

- One theoretical difference worth knowing: if `(consolidation_key, season)` ever had **duplicate rows** in `consolidated_owners_balances_memes`, the old single-key query took row `[0]`'s `sets` (even if 0 → 1), while the batch keeps the last truthy `sets`. That pair is the table's identity, so duplicates shouldn't exist — flagging for completeness.
- Ordering note: new auto-subscription rows are now appended in input order rather than `Promise.all` resolution order; both orders were semantically meaningless (rows are upserted; logs are inserts with auto-ids).
- Follow-up candidate spotted, out of scope: `consolidateSubscriptions` still queries the consolidation-keys view per wallet in a nested loop (small n — runs only for changed addresses) and builds a `LIKE '%...%'` filter by string concatenation, which is worth a look on its own.

## Summary by CodeRabbit

* **New Features**
  * Subscription eligibility and resulting counts are now calculated using batched lookups across multiple consolidation keys for better performance.

* **Bug Fixes**
  * Improved subscription counting for both automatic and final subscriptions, including cases with no available seasons and missing/empty eligibility data.
  * More consistent eligibility fallback behavior when eligibility isn’t found in the database.

* **Tests**
  * Added Jest coverage for batched eligibility lookup, key normalization and deduping, query chunking behavior, and default/fallback scenarios.

### DIFF
--- a/src/api-serverless/src/subscriptions/api.subscriptions.db.ts
+++ b/src/api-serverless/src/subscriptions/api.subscriptions.db.ts
@@ -36,7 +36,10 @@ import { BadRequestException } from '@/exceptions';
 import { getMaxMemeId } from '@/nftsLoop/db.nfts';
 import { sqlExecutor } from '@/sql-executor';
 import { equalIgnoreCase } from '@/strings';
-import { fetchSubscriptionEligibility } from '@/subscriptionsDaily/db.subscriptions';
+import {
+  fetchSubscriptionEligibility,
+  fetchSubscriptionEligibilityForKeys
+} from '@/subscriptionsDaily/db.subscriptions';
 import { Time } from '@/time';
 
 const SUBSCRIPTIONS_START_ID = 220;
@@ -819,18 +822,9 @@ async function fetchEffectiveUpcomingMemeSubscriptionCounts(
     balanceMap.set(b.consolidation_key.toLowerCase(), b.balance);
   });
 
-  // Fetch subscription eligibility for auto subscriptions
-  const autoSubEligibilityMap = new Map<string, number>();
-  await Promise.all(
-    autoSubs.map(async (autoSub) => {
-      const eligibility = await fetchSubscriptionEligibility(
-        autoSub.consolidation_key
-      );
-      autoSubEligibilityMap.set(
-        autoSub.consolidation_key.toLowerCase(),
-        eligibility
-      );
-    })
+  // Fetch subscription eligibility for auto subscriptions in one batched query
+  const autoSubEligibilityMap = await fetchSubscriptionEligibilityForKeys(
+    autoSubs.map((autoSub) => autoSub.consolidation_key)
   );
 
   const counts: SubscriptionCounts[] = [];
--- a/src/subscriptionsDaily/db.subscriptions.ts
+++ b/src/subscriptionsDaily/db.subscriptions.ts
@@ -115,9 +115,11 @@ export async function fetchSubscriptionEligibilityForKeys(
   consolidationKeys: string[]
 ): Promise<Map<string, number>> {
   const eligibility = new Map<string, number>();
+  const queryKeys = new Set<string>();
   consolidationKeys.forEach((key) => {
     if (key) {
       eligibility.set(key.toLowerCase(), 1);
+      queryKeys.add(key);
     }
   });
   if (eligibility.size === 0) {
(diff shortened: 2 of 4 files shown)


punk6529:
@coderabbitai review

PR #1704, opened 2026-07-04
Upgrade json-2-csv to v5

punk6529:
## Issue
- `json-2-csv@3.18.0` carries a moderate CSV-injection advisory (its `preventCsvInjection` option is ineffective in affected versions). It is the only remaining root audit finding fixable without touching the serverless v3 chain, and `npm audit fix` cannot apply it because the fix is a semver-major.
- Note the codebase never passes `preventCsvInjection`, so the flaw is not directly exploitable here — this clears the advisory and modernizes the API.

## Fix
- Bump to `json-2-csv@^5.5.11` and adapt the seven call sites to the v5 API.

## Changes
- `package.json` / `package-lock.json`: json-2-csv `^3.18.0` → `^5.5.11`.
- v5 renamed `json2csvAsync` → `json2csv` and made it synchronous. All call sites switched to named imports and dropped the now-meaningless `await`: `api-helpers.ts` (API CSV downloads), `dbDumpsDaily`, `rates/ratings.service.ts`, `rememesLoop`, `royalties.ts`, `subscriptionsDaily`, `tdhLoop/tdh_upload.ts`.

## Validation
- Output equivalence: v3 `json2csvAsync` and v5 `json2csv` produce byte-identical CSV for representative shapes (header derivation, comma quoting) and the empty-array edge case (both yield `"\n"`).
- Full root build + test suite green locally (all suites, 319s).
- eslint clean on all changed files.
- `npm audit`: the json-2-csv advisory is gone; remaining findings are exclusively the serverless-v3 chain (deferred — fix requires serverless 4, a semver-major with a commercial licensing change, so that is an owner decision rather than a mechanical bump).

## Risk
- Level: Low-Medium
- Why: touches CSV generation used by API download endpoints and several upload loops; mitigated by output-equivalence verification and the full test suite. The sync API change is absorbed at the call sites; error propagation inside existing try/catch blocks is unchanged.
- Rollback: revert the commit and redeploy affected services.

## Deployment
- `api`, `dbDumpsDaily`, `rememesLoop`, `royaltiesLoop`, `subscriptionsDaily`, `tdhLoop` (each bundles a changed file or the bumped dependency). Order-independent; staging `api` + `tdhLoop` serve as the primary validation pair. Undeployed loops keep running the old bundled v3 harmlessly until their next regular deploy.

## Review Notes
- v5 returns synchronously; `await` was removed rather than kept to avoid await-of-non-promise smell. Behavior inside callers is identical.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,7 +44,7 @@
         "image-size": "^2.0.2",
         "imagescript": "^1.3.1",
         "joi": "^17.11.0",
-        "json-2-csv": "^3.18.0",
+        "json-2-csv": "^5.5.11",
         "memory-cache": "^0.2.0",
         "merkletreejs": "^0.6.0",
         "moment-timezone": "^0.5.45",
@@ -11010,10 +11010,12 @@
       }
     },
     "node_modules/deeks": {
-      "version": "2.6.1",
+      "version": "3.2.1",
+      "resolved": "https://registry.npmjs.org/deeks/-/deeks-3.2.1.tgz",
+      "integrity": "sha512-D/o0k3pCG1aI1cxb/dDiWmtMc4Rh7ZQBybXpfMsw9Rbtqwg8kUA9SpYkWcw0pAUjZSnPm8MluctiS0o68r69jQ==",
       "license": "MIT",
       "engines": {
-        "node": ">= 12"
+        "node": ">= 16"
       }
     },
     "node_modules/deep-extend": {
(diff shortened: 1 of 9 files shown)


punk6529:
C:/Program Files/Git/6529bot review general security auth-api db-lambda

punk6529:
@6529bot review general security auth-api db-lambda

PR #1705, opened 2026-07-04
perf(indexing-loops): batch per-item queries and remove quadratic rank/count scans

punk6529:
## Issue

Four indexing/stats jobs issue one query (or one full-array scan) **per item** where one grouped query (or one precomputed map) serves the whole run:

- `nextgen_tokens.processTokens`: one `NextGenTokenTrait` query **per token** — ~10k queries per collection per metadata refresh — plus an O(tokens²) trait-count "denominator" rescan per token.
- `nftsLoop/nfts.updateSupply`: one `SUM(balance)` query **per ERC1155 NFT** per refresh, and a full-map rescan per ERC721 NFT to count same-contract entries.
- `nftsLoop/nft-extended-data.assignRanksByValue`: O(n²) rank assignment (count strictly-better per element), executed for ~6 rank fields per run.
- `marketStatsLoop`: one 4-window volume aggregation query **per NFT** per hourly stats run.

Part of a batch of performance-only PRs from a full backend audit; all changes intended to be **output-identical**.

## Fix

- **NextGen traits**: chunked `In(tokenIds)` fetch (5000/chunk) grouped into `Map<token_id, traits[]>`. Trait usage is order-independent (sums, products, `Math.min`, `Set` membership), so no `ORDER BY` dependency exists. Denominator = frequency map over per-token trait counts, built once — same counts as the old per-token filter. Queries still run on the same `entityManager` (transaction) connection, just ~10k× fewer round trips.
- **updateSupply**: one `GROUP BY contract, token_id` SUM over the ERC1155 contracts present, looked up per NFT (`?? 0` matches the old `Number(raw?.sum ?? 0)` for tokens with no owner rows); meme #8 burn adjustment unchanged. ERC721 branch reads a per-contract entry count computed once (the old filter counted **all** entries of the contract regardless of token type — the new count does exactly the same). Case-insensitivity of the old SQL `=` comparisons is preserved by lowercased map keys.
- **assignRanksByValue**: sort by `(value per direction, id asc)` and assign `index + 1`. Since ids are unique this is a total order, so "1 + number of strictly better items" is exactly the sorted position. **Property-tested against the verbatim old implementation** (200 randomized runs incl. duplicate values, both directions).
- **Market stats volumes**: one `GROUP BY token_id` query per contract with the same four conditional SUM windows. Tokens with no transactions are absent from the map → `volumes` is `undefined` → the existing `volumes?.x ?? 0` defaults produce the same zeros the old all-NULL single-row result did. Windows are now evaluated against a single `NOW()` per contract instead of a `NOW()` per NFT milliseconds apart — same-run consistency, no semantic change.

## Changes

- `src/nextgen/nextgen_tokens.ts` — chunked trait fetch + grouping; trait-count frequency map.
- `src/nftsLoop/nfts.ts` — grouped supply query + per-contract counts in `updateSupply`.
- `src/nftsLoop/nft-extended-data.ts` — sort-based `assignRanksByValue` (now exported for the test).
- `src/marketStatsLoop/nft_market_stats.ts` + `src/db.ts` — `findVolumesForContract` (grouped) replaces per-NFT `findVolume` in the loop; `findVolume` itself is kept (no other callers removed).
- `src/nftsLoop/assign-ranks-by-value.test.ts` — **new**: fast-check equivalence property vs the old counting implementation + explicit tie-break case.

## Validation

- New property test: 2/2 pass (200 randomized cases).
- Existing `src/nftsLoop/nft-extended-data.test.ts` (rank behavior through the public API): 5/5 pass unchanged.
- Not tested live: full nextgen/nfts/marketStats loop runs against production data.

## Risk

- Level: **Low**
- Why: batch jobs only; every rewrite mechanically preserves the old per-item semantics, with the only behaviorally-subtle one (rank assignment) property-tested against the old algorithm.

## Deployment

Not to be deployed yet — awaiting human review (@gelatogenesis). When approved: `nextgenContractLoop`, `nextgenMetadataLoop` (both import the nextgen module), `nftsLoop`, `marketStatsLoop`.

## Review Notes

- `assignRanksByValue` assumes rank-field values are real numbers (they're computed unconditionally upstream). If a value were ever `NaN`, the old counting code produced degenerate all-rank-1-ish results and the new sort produces engine-order-dependent results — both garbage; flagged only for completeness.
- In `updateSupply`, the ERC1155 supply map intentionally keys by `contract_tokenId` because MemeLab and Memes share token id ranges.

## Summary by CodeRabbit

* **Performance Improvements**
  * NFT stats, trait processing, and supply updates now use batched lookups, which should make large collections load and process faster.
* **Bug Fixes**
  * Improved rank calculations and tie handling for NFT sorting, making displayed rankings more consistent.
  * Volume and supply values are now computed more reliably across token batches.
* **New Features**
  * Expanded support for reusable NFT volume data handling behind the scenes, improving consistency in results.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -873,15 +873,17 @@ export async function persistMemesExtendedData(data: MemesExtendedData[]) {
   await AppDataSource.getRepository(MemesExtendedData).save(data);
 }
 
-export async function findVolume(
-  nft_id: number,
-  contract: string
-): Promise<{
+export interface NftVolumes {
   total_volume_last_24_hours: number;
   total_volume_last_7_days: number;
   total_volume_last_1_month: number;
   total_volume: number;
-}> {
+}
+
+export async function findVolume(
+  nft_id: number,
+  contract: string
+): Promise<NftVolumes> {
   const sql = `SELECT
       SUM(CASE WHEN created_at >= DATE_SUB(NOW(), INTERVAL 24 HOUR) THEN value ELSE 0 END) AS total_volume_last_24_hours,
       SUM(CASE WHEN created_at >= DATE_SUB(NOW(), INTERVAL 7 DAY) THEN value ELSE 0 END) AS total_volume_last_7_days,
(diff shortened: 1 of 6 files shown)


punk6529:
@coderabbitai review

punk6529:
@coderabbitai review

PR #1707, opened 2026-07-04
perf(nextgen): remove quadratic trait-score scans in collection processing

punk6529:
## Issue

Tier-1 follow-up to the perf batch under review (#1699, #1701, #1703, #1705, #1706) — this one is likely the **largest remaining single win** in the codebase.

`processCollectionTraitScores` in `src/nextgen/nextgen_tokens.ts` runs on every NextGen contract/metadata loop pass and, for **every trait row**, re-filters the entire trait-row list of the collection (then scans that subset twice more). With T = tokens × trait categories ≈ 150k rows for a 10k-token collection, that's on the order of 10⁹–10¹⁰ operations per run. `calulateTokenRanks` adds a full-list filter per trait category per rank field (6 fields × ~dozens of categories × 150k).

## Fix

- **Per-trait value-count map built once** (`Map<trait, Map<value, count>>`). Per row, `trait_count` (= distinct values for the trait), `value_count` / `sharedKeyValue` (= rows with the same trait+value) and `valuesCountForTrait` are O(1) lookups. Equivalence: the old code used strict `===` on `trait` and `value`; `Map`/`Set` keying is SameValueZero, which is identical to `===` for the string (and even null) values involved. The old code also recomputed `sharedKeyValue`/`valuesCountForTrait` redundantly per row — they are by definition `value_count`/`trait_count`, which the old code had already computed from the same subset.
- **`calulateTokenRanks` groups rows by category in one pass** per call instead of `Set` + per-category `filter`. Map insertion order ≡ the previous first-appearance category order, and per-category rows keep array order — so the inputs to the (untouched) `calculateTokenRanksForCategory` sorts are byte-identical, and therefore the ranks **and the output array order** are unchanged. This matters because the six chained calls deliberately reorder the array; the property test asserts order equality, not just content.

## Changes

- `src/nextgen/nextgen_tokens.ts` — the two rewrites above; `calculateTokenRanksForCategory` untouched.
- `src/nextgen/nextgen-token-trait-scores.test.ts` — **new**: fast-check property through the public `refreshNextgenTokens` entry proving the rows handed to `persistNextGenTraits` are deep-equal **including order** to the verbatim pre-optimization pipeline (kept in the test as a reference implementation), across randomized token/trait/value combinations including mint-type traits; plus an explicit mint-type case (−1 scores, counts still populated).

## Validation

- New tests: 2/2 pass (property runs 40 randomized collections).
- Not tested live: a full nextgen loop run against a production-size collection.

## Risk

- Level: **Low**
- Why: batch loop only; pure-computation rewrite property-tested for exact output equality including array order.

## Deployment

Not to be deployed yet — same review gate as the open perf batch (@gelatogenesis). When approved: `nextgenContractLoop`, `nextgenMetadataLoop`.

## Review Notes

- This file is also touched by open PR #1705, but in a **different function** (`processTokens`) — the hunks don't overlap, so either merge order is clean.
- The interesting equivalence detail is the array-order preservation through the six chained `calulateTokenRanks` calls (explained above; asserted by the test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token collection trait score calculations with more consistent rarity values.
  * Enhanced handling for mint-type trait rows so rarity-related scores are correctly set, while counts remain accurate.
  * Improved ranking stability by preserving category ordering based on first appearance.

* **Tests**
  * Added property-based and deterministic tests to verify trait-score persistence matches expected pipeline output (including correct row ordering).

### DIFF
--- a/src/nextgen/nextgen-token-trait-scores.test.ts
+++ b/src/nextgen/nextgen-token-trait-scores.test.ts
@@ -1,0 +1,247 @@
+import fc from 'fast-check';
+import { NextGenCollection, NextGenTokenTrait } from '@/entities/INextGen';
+import {
+  fetchNextGenCollections,
+  fetchNextGenTokensForCollection,
+  fetchNextGenTokenTraits,
+  persistNextGenCollection,
+  persistNextGenCollectionHodlRate,
+  persistNextGenTraits,
+  persistNextGenTraitScores
+} from './nextgen.db';
+import { MINT_TYPE_TRAIT } from './nextgen_constants';
+import { refreshNextgenTokens } from './nextgen_tokens';
+
+jest.mock('./nextgen.db', () => ({
+  fetchNextGenCollections: jest.fn(),
+  fetchNextGenTokensForCollection: jest.fn(),
+  fetchNextGenTokenTraits: jest.fn(),
+  persistNextGenCollection: jest.fn(),
+  persistNextGenCollectionHodlRate: jest.fn(),
+  persistNextGenTraits: jest.fn(),
+  persistNextGenTraitScores: jest.fn()
+}));
+
+const mockedFetchNextGenCollections =
+  fetchNextGenCollections as jest.MockedFunction<
+    typeof fetchNextGenCollections
+  >;
+const mockedFetchNextGenTokensForCollection =
+  fetchNextGenTokensForCollection as jest.MockedFunction<
+    typeof fetchNextGenTokensForCollection
+  >;
+const mockedFetchNextGenTokenTraits =
+  fetchNextGenTokenTraits as jest.MockedFunction<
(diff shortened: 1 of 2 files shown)


punk6529:
@coderabbitai review

punk6529:
@coderabbitai review

punk6529:
Re the nitpick about `any` types in the test's reference helpers: keeping them loosely typed is deliberate. The reference oracle is a verbatim copy of the pre-optimization production code (which itself used `as any[]` and dynamic ``tt[`${field}_rank`]`` access), and its value comes from being character-for-character the old algorithm. A misspelled field in the oracle can't pass silently: the property compares deep equality against the real pipeline's output, so any divergence fails the test loudly. Tightening the types would mean editorializing the copy, which weakens the equivalence argument.

PR #1708, opened 2026-07-04
perf(consolidations): batch consolidation-key lookups in activity and nft-owners loops

punk6529:
## Issue

Tier-1 follow-up to the perf batch under review. `consolidateActivity` (aggregatedActivityLoop) and `consolidateNftOwners` (nftOwnersLoop) contain the **exact same pattern** fixed for ownersBalancesLoop in #1701:

- one `address_consolidation_keys` view query **per address** inside an unbounded `Promise.all`, and
- the consolidated aggregate recomputed **once per member address** of the same consolidation (members overwrite each other's identical results).

Both functions also run inside `delegationsLoop` on every consolidation change, so the win applies to three services.

## Fix

New shared helper `fetchConsolidationGroupsForAddresses(addresses)` in `src/db.ts`:

- resolves all addresses through the view in **5000-address chunked queries**;
- returns one entry per unique consolidation: `Map<consolidation key, member addresses>`;
- preserves the old per-address semantics exactly: address missing from the view → singleton group keyed by the lowercased address (the old `[0]`-or-fallback branch); first row wins per address; group order follows first appearance in the input.

Both call sites now iterate unique groups and compute each consolidation's aggregate **once**. Equivalence: same-group members previously invoked the aggregate with identical `(key, key.split('-'))` arguments and overwrote the same map entries with identical values, and the delete-delta was already the union of group members — so map contents and delta are unchanged.

## Changes

- `src/db.ts` — `fetchConsolidationGroupsForAddresses` (+ chunk-size constant).
- `src/aggregatedActivityLoop/aggregated_activity.ts` — `consolidateActivity` uses the helper; per-group activity + memes-activity computed once.
- `src/nftOwnersLoop/nft_owners.ts` — `consolidateNftOwners` uses the helper; one owners fetch per unique group.
- `src/fetch-consolidation-groups-for-addresses.test.ts` — **new**: dedup + fallback, first-appearance ordering, 5000-chunking, empty-input no-query.
- `src/nftOwnersLoop/consolidate-nft-owners.test.ts` — **new**: one owners fetch per unique group, member balance summing, delete-delta union, empty-set no-op.

## Validation

- New tests: 5/5 pass.
- Not tested live: full loop runs against production data.

## Risk

- Level: **Low**
- Why: mechanical application of the already-reviewed #1701 pattern, now via one shared, unit-tested helper.

## Deployment

Not to be deployed yet — same review gate as the open perf batch (@gelatogenesis). When approved: `aggregatedActivityLoop`, `nftOwnersLoop`, `delegationsLoop` (calls both functions).

## Review Notes

- The helper reads the view row's `address` column through a local cast because `WalletConsolidationKey` still declares the misnamed `wallet` field on `main`; open PR #1701 fixes that interface, after which the cast can be dropped in a tiny follow-up. Kept the cast here so the two PRs don't both edit `IDelegation.ts`.
- ownersBalancesLoop keeps its own inline copy of this logic from #1701 for now; once both PRs land, pointing it at the shared helper is a two-line cleanup (happy to do as a follow-up).

## Summary by CodeRabbit

- **New Features**
  - Consolidation now performs batched grouping of addresses before computing consolidated activity, reducing per-address lookups.
  - NFT owner consolidation now runs per unique consolidation group to improve performance on large address sets.

- **Bug Fixes**
  - Addresses without a consolidation mapping are now handled as singleton groups.
  - Ensures stable group ordering based on first appearance and correctly skips work for empty inputs.

- **Tests**
  - Added Jest coverage for batched, chunked grouping behavior (including 5,000-item batches) and ordering.
  - Updated NFT owner consolidation tests to validate grouped fetch and persistence behavior.

### DIFF
--- a/src/aggregatedActivityLoop/aggregated_activity.ts
+++ b/src/aggregatedActivityLoop/aggregated_activity.ts
@@ -11,9 +11,9 @@ import {
 import { Transaction } from '../entities/ITransaction';
 import {
   fetchAllSeasons,
+  fetchConsolidationGroupsForAddresses,
   fetchMaxTransactionsBlockNumber,
   fetchTransactionsAfterBlock,
-  fetchWalletConsolidationKeysViewForWallet,
   fetchWalletTransactions
 } from '../db';
 import { Logger } from '../logging';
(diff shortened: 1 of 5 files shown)


punk6529:
@coderabbitai review

punk6529:
@coderabbitai review

PR #1717, opened 2026-07-06
Allow dashes in rep categories (never leading) with rule-specific errors

brookr:
## Issue
- Rep category names rejected the dash, forcing awkward names (`hey-jude`, `Solidity-Dev` were impossible) — the allowed separator set was very tight.
- When rejected, users got one catch-all blob listing every rule at once (see the Grant Rep modal screenshot in the wave discussion) instead of the specific rule they broke.

## Fix
- `REP_CATEGORY_PATTERN` now accepts the ASCII dash anywhere **except the first character**: `^[\p{L}\p{N}?!,.'() ][\p{L}\p{N}?!,.'() -]{0,99}$` (total length still 1–100).
  - Leading dash stays banned deliberately: it's a spreadsheet formula-injection trigger (`-`, like `=`/`+`/`@`) for CSV/open-data exports, and reads as a negative sign next to signed rep amounts. The rationale is commented at the pattern.
  - Unicode en/em dashes stay banned (identity-key homoglyphs — categories aggregate per exact string).
- New `explainRepCategoryViolation()` (`src/profiles/rep-category-rules.ts`) gives the **exact rule violated**: empty / over-length with the actual count / leading dash / the specific disallowed characters used (named, once each, invisible chars described in words). Wired into both abusiveness pre-checks (single + bulk).
- The identical validation message that was copy-pasted across six route files is now one exported constant, updated to state the dash rule clearly.
- OpenAPI schemas (7 pattern occurrences) updated; `restructure-openapi` + `generate` run (no generated-model diffs — patterns are validation-only).

## Changes
- `src/entities/IAbusivenessDetectionResult.ts` — pattern + rationale comment + shared message constant.
- `src/profiles/rep-category-rules.ts` (new) + wiring in `abusiveness-check.service.ts`.
- Six Joi sites now use the shared constant (`profile-rep`, `wave-rep`, `ratings`, `bulk-rep`, `waves`, `global-rep-category` — the last also gains the friendly message it lacked).
- Tests (new, 49 cases across two suites): leading-dash misuse (incl. SQL-comment lookalikes, fake negative amounts), spreadsheet formula triggers, separator/structure chars still banned (colon, slash, etc.), unicode dash homoglyphs, length bounds, plus fast-check properties: backward compatibility (everything valid before stays valid), no accepted string starts with a dash, and **explainer ⟷ pattern equivalence on arbitrary unicode** (lengths counted in code points to match the pattern's `u`-flag semantics).

## Validation
- `npm test` on both new suites: 49/49. `npm run lint`: clean. Root `tsc --noEmit`: clean.
- Existing categories are provably unaffected (backward-compat property test) — no data migration needed.

## Risk
- Level: Low
- Why: strictly widens accepted input by one character in non-leading positions; every downstream consumer (URLs, AI check, notifications, exports) was audited for dash-safety before this change. Error-message changes are user-facing text only.
- Rollback: revert the commit (already-created dash categories would then fail validation on *new* writes but remain readable).

## Deployment
- `api` lambda only (all rep-category validation runs in the API; no loop consumes the abusiveness pre-check).

## Review Notes
- FE has no client-side mirror of this rule — it surfaces the server message, so this PR is the full fix for acceptance. A follow-up FE PR will improve how the Grant Rep modal *presents* these errors (a11y/i18n/responsive).
- Architecture docs: no update needed — no system-shape change (validation + messages only).

## Summary by CodeRabbit

* **New Features**
  * Added clearer, more specific validation messages for REP category inputs.
  * Improved category rules so inputs now enforce allowed characters, length limits, and dash placement more consistently.

* **Bug Fixes**
  * Standardized REP category validation across profiles, ratings, waves, and category flows.
  * Fixed apostrophe text rendering in several API descriptions for better readability.


brookr:
Fixed in the follow-up commit: `openapi.yaml` is restored to main's exact text with only the seven rep-category `pattern` lines changed — the apostrophe corruption (`user's` → `user\"s`) and the unrelated `ApiIdentityMuteState` reordering came from `npm run restructure-openapi`'s YAML round-tripping, not hand edits. Filed the tool bug separately so other PRs don't trip on it. `npm run generate` output is unaffected (patterns don't emit into models), which PR CI verifies.

brookr:
Good catch on the second pattern family — fixed: the five ASCII-only `category` param patterns (`getRepRatingByCategory` and friends) now carry the same dash rule (`^[a-zA-Z0-9?!,.\"() ][a-zA-Z0-9?!,.\"() -]{0,99}$`), so a dash category round-trips through read endpoints. Their backing Joi validators already used the shared `REP_CATEGORY_PATTERN` updated in the first commit, so this was documentation-pattern drift only — but it would have bitten any client that pre-validates against the spec.

brookr:
Re the GLM swarm advisories:

- **"Cannot verify never-leading from the diff slice"**: the definitions are in this PR's first commit — `REP_CATEGORY_PATTERN` anchors the first character to the dash-free class (`^[\p{L}\p{N}?!,.'() ][\p{L}\p{N}?!,.'() -]{0,99}$`, `src/entities/IAbusivenessDetectionResult.ts`) and `REP_CATEGORY_INVALID_MESSAGE` states the rule. Every test the advisory asks for already exists in `src/entities/rep-category-pattern.test.ts`: leading dash (`-Unreliable`, `-`, `--`), embedded (`Solidity-Dev`), multiple (`double--dash`), empty, 101 chars, pre-existing punctuation — plus fast-check properties for backward compatibility and no-accepted-string-starts-with-dash.
- **Screening consistency**: verified — every rep-category-accepting route reaches the abusiveness check: `ratings.routes` and `profile-rep`/`wave-rep` call `checkRepPhrase` directly; `bulk-rep` screens inside `ratingsService.bulkRep` → `bulkCheckRepPhrases` (`src/rates/ratings.service.ts:1311`). No unscreened path.
- Route-level 400 tests per endpoint are a fair suggestion but need the testcontainers harness; deferring as follow-up rather than expanding this validation-scoped PR.

brookr:
Review round addressed (267ec7b2 + c4af975b):

- **openapi:5726 (remaining schemas)**: fixed — `ApiBulkRateRequest.category` carried a third, quantifier-less pattern variant (`^[...]$`, single char — pre-existing spec bug) that both replacement passes missed. All **13** category patterns in the spec now share the dash rule.
- **openapi:2061 (apostrophes)**: stale — that inline was reviewed against the pre-fix commit; the current head has main's exact text (`user's` etc., zero double-quote corruptions). Verified by grep.
- **Alias imports**: done for both lines this PR authored (test + service).
- **Regex literals**: attempted, then reverted with documentation — `/.../u` literals are a **TS1501 compile error at the repo's es5 target**; the constructor form is load-bearing, not style. Both sites now carry a comment so the suggestion doesn't resurface.
- **Duplicated allowed-chars summary**: leaving as-is — the Joi constant is a complete user-facing sentence while the explainer composes rule-specific fragments; importing one into the other would couple entities→profiles or invert the existing dependency direction.

49/49 tests, lint clean, root tsc clean on the current head.

GelatoGenesis (approved):

PR #1739, opened 2026-07-08
feat(eligibility): add dark membership materialization schema

punk6529:
## What

Adds the schema foundation for eligibility membership materialization. It remains behavior-neutral on its own; the refresh loop, producer hooks, backfill, and read path land in the implementation PR.

- **`user_group_members`**: materialized criteria-group membership with a composite primary key, profile-to-groups index, rule-spec version, and input watermark.
- **`membership_refresh_requests`**: durable, coalescing PROFILE/GROUP dirty markers with retry metadata.
- **`membership_materialization_states`**: successful target refresh state, including authoritative empty profile/group results.
- **`membership_watermarks`**: global/per-dimension materialization watermarks and future chain-height detail.

All entities are exported through `entities.ts`. Per repository policy, schema synchronization is handled by `dbMigrationsLoop`: no migration files, foreign keys, or database-level enum constraints are introduced.

The branch is current with `main`.

## Tests

- `npm test -- src/entities/membership-schema.db.test.ts --runInBand`
- 1 suite, 4 tests passed
- DCO signoff present

## Architecture

`docs/architecture.md` is intentionally unchanged here because this PR adds schema only. The implementation PR documents the new Lambda, SQS queues, producer flow, materialization state, and read fallback.

## Deployment

Deploy `dbMigrationsLoop` before the membership refresh implementation so TypeORM sync creates all four tables.

## Summary by CodeRabbit

* **New Features**
  * Added support for persisting user-group membership records.
  * Added persistence for membership refresh requests, including retry and error tracking.
  * Added storage for membership materialization state to reflect sync progress.
  * Added membership watermark persistence to track per-dimension progress.
  * Exposed the new membership data models for application use.
* **Tests**
  * Added database round-trip coverage for the membership schema tables, including default values and nullable fields.


GelatoGenesis (approved):

GelatoGenesis:
Addressed the review in `65110950b`:

- Verified the pinned TypeORM 0.3.31 MySQL implementation: `MysqlQueryRunner.createTable` calls `createTableSql`, which inlines secondary indexes in the CREATE statement. The DB test now explicitly checks that every declared index is in the seven CREATE statements and verifies the persisted indexes. The create-only guard stays narrow: separate index/ALTER repair of an existing table is deliberately rejected. `downQueries` describe reversal and are never executed by this operation.
- Added a fault-injection test that fails the fourth CREATE. The successful prefix remains; retry creates exactly the four missing tables and then verifies all seven. A further invocation is a no-op. The design now documents this MySQL autocommit recovery path; a transaction cannot make the seven DDL statements atomic.
- Confirmed fixture isolation: globalSetup starts Testcontainers MySQL and provisions one database per Jest worker. describeWithSeed resets that worker's database. No shared development/staging database is used.
- The wave-score entity deliberately preserves current main, as requested for this additive foundation. It is outside this schema scope; the test preserves its full table definition and row values. No legacy typing refactor is included.

Validation: the expanded schema, handler and deployment-guard suites pass (84 tests), as does lint. The additional isolated MySQL 8.4.8 run, matching the staging RDS engine version, also passed all 11 schema tests, including inline-index creation, fourth-CREATE failure recovery and idempotence. The temporary Testcontainers configuration enables mysql_native_password for the repository's existing MySQL client; no shared database was used.

PR #1748, opened 2026-07-13
Count unread drops across all subwaves

GelatoGenesis:
## Issue
- Parent wave unread badges only counted followed subwaves, so the total beside the subwave control could disagree with the visible child unread badges.

## Fix
- Add an additive aggregate that counts unread drops across every visible subwave with reader state, independent of follow status.
- Preserve the legacy followed-only fields for a zero-downtime backend-first rollout.

## Changes
- Extend the wave overview profile context with `subwave_unread_drops`.
- Count unread drops using the same self-author, muted-profile, muted-wave, visibility, and reader-baseline rules as individual wave unread state.
- Return aggregate context even when no child subwave is followed.
- Add database and mapper regression coverage for mixed followed/unfollowed children and fully unfollowed child sets.
- No database migration or architecture-document update is needed.

## Validation
- `npm run lint`
- `npm run build` (226 suites, 2,027 tests passed)
- Focused wave database regression tests (2 passed)
- `cd src/api-serverless && npm run restructure-openapi && npm run generate && npm run build`

## Risk
- Level: Medium
- Why: additive API field with a broader unread aggregate query; no schema or destructive contract change.
- Rollback: redeploy the prior `api` version; the legacy followed-only fields remain intact.

## Deployment
- Deploy `api` only.
- Deploy backend before the corresponding frontend change in staging and production.

## Review Notes
- Please focus on unread parity with individual child counts and query behavior for parents with no followed children.

### DIFF
--- a/src/api-serverless/openapi.yaml
+++ b/src/api-serverless/openapi.yaml
@@ -19302,6 +19302,12 @@ components:
         first_unread_drop_serial_no:
           type: integer
           format: int64
+        subwave_unread_drops:
+          type: integer
+          format: int64
+          description: >-
+            Unread drops across all visible subwaves, independent of whether the
+            authenticated profile follows each subwave.
         followed_subwaves_count:
           type: integer
           format: int64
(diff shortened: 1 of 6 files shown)


GelatoGenesis:
Addressed the actionable review points in e4bc956e:

- Aligned the legacy followed-only unread count and first-unread serial with normal unread semantics, so self-authored and muted-author drops are excluded from both aggregates.
- Added explicit self-authored, muted-author, followed, unfollowed, and no-follow regression coverage.
- Clarified that this path has no persistent cache: it only coalesces equivalent requests until the database query settles, so a read or mute change is visible on the next request.

I kept `subwave_unread_drops` optional in OpenAPI intentionally. The runtime always emits it on the new backend, but optional typing preserves compatibility for mixed-version rollout and backend rollback while the frontend uses the legacy fallback.

PR #1751, opened 2026-07-13
Add wave-scoped competition badge API

GelatoGenesis:
## Issue
- Wave chats need participant and winner badges scoped to the current competition without inflating every drop response with full entry data.

## Fix
- Add compact wave-scoped participation booleans to standard drop authors and a separate paginated endpoint that returns full competition drops only when requested.
- Use indexed EXISTS probes so badge metadata remains bounded and does not enumerate entry IDs.

## Changes
- Add `is_participant` and `is_winner` author metadata outside Main Stage.
- Add `GET /v2/waves/{id}/competition-drops` with author, drop-type, page, and page-size validation.
- Enforce wave and parent-wave visibility before returning entries.
- Regenerate OpenAPI models and routes.
- Add DB, mapper, service, and handler coverage.
- No schema migration or loop deployment is required; only the `api` service changes.

## Validation
- `npm run lint`
- `TESTCONTAINERS_RYUK_DISABLED=true npm run build`
- 226 suites passed; 2,034 tests passed and 1 skipped.
- Local query benchmark across 100 drops measured about 0.87 ms median and 1.15 ms p95 for badge flags.

## Risk
- Level: Medium
- Why: This adds bounded read work to drop mapping and a new read-only API endpoint, but preserves existing fields and requires no migration.
- Rollback: Revert this PR and redeploy the API service. The frontend must be deployed only after this backend reaches staging.

## Review Notes
- Focus on visibility enforcement, indexed query shape, Main Stage omission, and pagination.
- Companion frontend PR will consume the additive contract after backend staging is healthy.

## Summary by CodeRabbit

* **New Features**
  * Added `GET /v2/waves/{id}/competition-drops` (V2) to retrieve paginated competition drops by author and drop type, including `page`/`page_size`.
  * Extended identity responses with wave participation details (participant/winner) for competition-related drops.
  * Added request validation for supported drop types and pagination parameters.
* **Tests**
  * Added coverage for the new endpoint, pagination behavior, DB query logic, and wave participation mapping.


GelatoGenesis:
Visibility finding confirmation: `ApiDropMapper.mapDrops` enriches only drop entities that the calling feed/endpoint has already authorized for the current viewer. The participation lookup is display-only and is never used to grant access. A caller who cannot see a wave cannot obtain one of its mapped drop contexts, while `GET /v2/waves/{id}/competition-drops` independently calls `assertWaveAndParentVisibleOrThrow` before returning any full entries. I documented this invariant beside the mapper lookup. The latest follow-up also treats a promoted `WINNER` row as both participant and winner and adds regression coverage.

PR #1752, opened 2026-07-14
Fallback to Sharp for unsupported GIF resizing

prxt6529:
## Issue
- The `s3Uploader` Lambda cannot resize some otherwise valid animated GIFs when ImageScript's WebAssembly decoder throws `RuntimeError: unreachable`.
- Resize errors were swallowed, causing the SQS record to be acknowledged even though derivative images were missing.

## Fix
- Keep ImageScript as the primary GIF resizer to preserve existing behavior.
- Fall back to Sharp with animated-image support only when ImageScript fails.
- Propagate failures when both resizers fail so the SQS partial-batch response can retry or redrive the record.

## Changes
- Added the animated Sharp fallback and a targeted fallback log entry.
- Added focused tests for the existing ImageScript path, Sharp fallback, and double-failure propagation.
- No API, database, entity, configuration, or architecture changes.

## Validation
- Focused image resize tests pass.
- Lint passes with zero warnings.
- Repository TypeScript compilation passes after installing root and API lockfile dependencies.
- Verified against the affected 24-frame, 1080x1350 production GIF: ImageScript reproduces `unreachable`; Sharp produces a valid 24-frame GIF at the requested height.
- The full all-tests build could not start its MySQL-backed suite because no container runtime is available in the local environment.

## Risk
- Level: Low
- Why: Existing successful ImageScript behavior is unchanged; the new code runs only after that path throws. Sharp is already bundled by `s3Uploader`.
- Rollback: Revert this PR to restore ImageScript-only GIF resizing.

## Deployment
- Redeploy `s3Uploader` only.

## Review Notes
- Please focus on animated GIF preservation and failure propagation to the SQS partial-batch handler.


prxt6529:
Addressed the actionable review feedback in `54a720e8`:

- Tightened the local `resizeImage` contract to `Promise<Buffer>`; its only call site in `src/s3.ts` intentionally propagates terminal resize failures to the existing SQS partial-batch handler.
- Added coverage proving the direct Sharp WebP path remains unchanged.
- Added an assertion that the ImageScript-to-Sharp fallback warning includes the original decoder failure.
- Reconfirmed the fallback is present in the complete PR diff and that the queue has `maxReceiveCount: 10` with a DLQ.

The broad fallback catch remains intentional: decode, resize, and encode incompatibilities are all eligible for the alternate decoder, while a failure from Sharp still propagates. Sharp's built-in input-pixel limit remains in effect, and the worker has bounded concurrency and Lambda memory.

PR #1770, opened 2026-07-20
Wire GitHub App identity into production API

GelatoGenesis:
## What changed

- resolve the Release Bus GitHub App installation for production API deployments
- inject the App ID and installation ID into `seizeAPI`
- validate the App configuration before deploying
- remove the production-only identity from staging API configuration
- add an infrastructure contract test covering the wiring

## Why

The Release Bus pause/resume endpoints returned HTTP 500 because `seizeAPI` loaded the private key but had no `RELEASE_BUS_GITHUB_APP_ID` or `RELEASE_BUS_GITHUB_INSTALLATION_ID`. The request failed before reaching GitHub or checking operator membership.

## Impact

Production API redeployment will make operator pause/resume authorization able to mint a GitHub App installation token. No database or public API schema changes are involved.

## Validation

- `npm run lint`
- `npm run build` (253 suites, 2,234 tests passed; TypeScript compilation passed)
- generated workflow parsed as YAML
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Production API deployments now support Release Bus integration through GitHub App credentials.
  * Release Bus installation details are automatically configured during eligible production deployments.

* **Bug Fixes**
  * Added validation to catch missing or incorrectly formatted Release Bus credentials before deployment.
  * Non-production deployments now reliably disable Release Bus integration and remove production-only settings.
  * Improved deployment configuration checks to prevent incomplete runtime setups.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -188,11 +188,15 @@ jobs:
         if: github.event.inputs.service == 'api' && github.event.inputs.environment == 'prod'
         shell: bash
         env:
+          RELEASE_BUS_GITHUB_APP_ID: ${{ vars.RELEASE_BUS_GITHUB_APP_ID }}
+          RELEASE_BUS_GITHUB_PRIVATE_KEY: ${{ secrets.RELEASE_BUS_GITHUB_PRIVATE_KEY }}
           RELEASE_BUS_GITHUB_WEBHOOK_SECRET: ${{ secrets.RELEASE_BUS_GITHUB_WEBHOOK_SECRET }}
           RELEASE_BUS_WORKFLOW_AUTH_TOKEN: ${{ secrets.RELEASE_BUS_WORKFLOW_AUTH_TOKEN }}
         run: |
           set -euo pipefail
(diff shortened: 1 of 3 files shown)


GelatoGenesis:
Addressed the review defensively in b8943c00: `RELEASE_BUS_MODE` is already inherited from workflow-level `env`, but the API deploy step now also declares it explicitly and validates the App ID immediately before production mutation. The generated workflow was regenerated from `scripts/generate-deploy-config.mjs`; `pretest`, `prelint`, and `prebuild` all regenerate it before validation. Targeted infrastructure tests and lint pass after the update.

GelatoGenesis:
Fixed the follow-up blocker in 4cd8d34c. The prior patch matched the wrong adjacent `env` block: the duplicate Release Bus mode entry is removed, `RELEASE_BUS_MODE` is now explicitly scoped to the `Deploy API` step, and App ID/installation ID validation runs before `update-function-code`. I also added a regression assertion against duplicate mode keys and verified the generated workflow parses with unique YAML keys. Generator, lint, targeted tests, and `git diff --check` are green.

PR #1772, opened 2026-07-20
Restore CI wave alerts after early deploy failures

prxt6529:
## Issue
- Early deploy validation failures occur before the repository checkout step.
- The Discord failure action remains available, but the repository-local `scripts/notify-ci-wave.mjs` file is absent, so the Waves notification cannot run.

## Fix
- Add a failure-only recovery checkout when the notifier script is missing.
- Check out the trusted workflow commit with persisted credentials disabled and sparse checkout limited to the notifier script.

## Changes
- Update the deploy workflow generator.
- Regenerate `.github/workflows/deploy.yml`.

## Validation
- Regenerated the deploy workflow successfully.
- Verified the generator with `node --check`.
- Parsed the generated workflow as YAML.
- Verified the diff with `git diff --check`.
- Full lint was not completed because the clean worktree lacked dependencies and the locked install remained silent; it was stopped rather than delaying publication.

## Risk
- Level: Low
- Why: The added checkout runs only after a failure and only when the notifier script is unavailable.
- Rollback: Revert this commit to restore the previous notification behavior.

## Deployment
- None. This changes the GitHub Actions workflow only.

## Review Notes
- Confirm that the failure-only sparse checkout remains pinned to `github.workflow_sha` and does not persist credentials.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deployment workflow reliability when failures occur early, ensuring completion/failure notifications still run.
  - Notification steps now automatically recover when the required notification script isn’t available after an early termination.

- **Chores**
  - Updated deployment automation to conditionally fetch only the notification script when needed.
  - Applied the same recovery behavior to generated deployment configurations.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -599,9 +599,6 @@ jobs:
           done
           echo "API did not report healthy exact commit $INPUT_EXPECTED_SHA" >&2
           exit 1
-      # Early validation failures skip the normal checkout. The remaining local
-      # failure notifier needs only this script; Discord uses an external action.
-      # Pin it to the trusted running workflow rather than the deploy target ref.
       - name: Checkout CI wave notifier after early failure
         if: failure() && hashFiles('scripts/notify-ci-wave.mjs') == ''
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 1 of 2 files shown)


prxt6529:
Verified both carried-forward leads:

- The recovery checkout runs only when `scripts/notify-ci-wave.mjs` is absent. The downstream failure notifications are the external Discord action and the Waves notifier, so no other repository files are required.
- `.github/workflows/on-pull-request.yml` already regenerates `deploy.yml` and fails when the committed generated file differs.

No additional code change is needed.

PR #1782, opened 2026-07-22
Complete Release Bus fast path and phase feedback

GelatoGenesis:
## Summary
- reuse exact trusted base evidence and exact-content backend preflight evidence fail closed
- deploy independent backend DAG frontiers concurrently while preserving dependency edges and sibling evidence
- expose current candidate phases, terminal failure details, and exact immutable SHA versus moved branch heads
- classify compose publication authorization/network failures as bounded infrastructure retries using the original composition artifact
- retain lane-local E2E infrastructure recovery and immutable deployment provenance

## Validation
- changed-file gate: 16 suites and 214 tests passed, lint and TypeScript passed
- focused worker/API gate: 8 suites and 163 tests passed
- API bundle build passed

Operational deployment will use release_note_publish=false.


GelatoGenesis:
Follow-up verification for the automated findings:

- The shared preflight npm activation is now single-quoted and contract-tested.
- The pinned official `actions/upload-artifact` commit explicitly declares `artifact-digest` in its `action.yml`: https://github.com/actions/upload-artifact/blob/ea165f8d65b6e75b540449e92b4886f43607fa02/action.yml#L49-L51. No version bump is needed.
- `advanceBackendDeploy` already inspects the `Promise.allSettled` frontier result and rethrows `rejectedDispatch` before emitting the frontier event; focused injected-rejection tests pass.
- The expected release-group service set is now persisted once at `:expected`; every later delivery must match it exactly before completion can be evaluated.

PR #1793, opened 2026-07-23
fix: accept Release Bus App workflow actors

GelatoGenesis:
## Summary
- recognize GitHub App `[bot]` logins when binding exact workflow-run identity
- retain the existing GitHub human-login length constraint
- cover both actor forms and invalid bracketed/overlong values

## Live beta evidence
The paused v2 staging beta dispatched compose run `29981702843` as `6529-release-bus[bot]`. The workflow stopped at its authorization boundary with `GitHub workflow run has no valid actor`; no shared ref or environment was mutated. ALL and STAGING remain paused.

## Validation
- `npm test -- --runInBand src/releaseBus/release-bus.github-app.test.ts src/releaseBusV2/release-bus-v2.operations.test.ts` (21/21)
- targeted ESLint
- `git diff --check`

Release notes intentionally disabled for this control-plane hotfix.

### DIFF
--- a/src/releaseBus/release-bus.github-app.test.ts
+++ b/src/releaseBus/release-bus.github-app.test.ts
@@ -121,8 +121,47 @@ describe('GitHub workflow operation identity', () => {
   it('accepts human and GitHub App workflow actors', () => {
     expect(isValidGitHubWorkflowActor('GelatoGenesis')).toBe(true);
     expect(isValidGitHubWorkflowActor('6529-release-bus[bot]')).toBe(true);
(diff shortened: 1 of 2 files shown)


GelatoGenesis:
Addressed the advisory in b3d789df: automated workflow identity is now limited to the exact live `6529-release-bus[bot]` actor, unrelated bots are rejected, and a mocked full `getWorkflowRunIdentity` path now covers the App actor. Targeted suites are 22/22 green; the prior full suite was 2513/2513 green.

PR #1795, opened 2026-07-23
fix: install API dependencies in v2 test shards

GelatoGenesis:
## Summary
- install frozen `src/api-serverless` dependencies in all API-aware typecheck/test shards
- include the nested lockfile in the setup-node npm cache key
- assert the dependency boundary in the v2 offline acceptance harness

## Live paused-beta evidence
Combined preflight run `29983424916` failed only because shard 4 could not resolve `serverless-http`; the other 505 tests in that shard passed. ALL/STAGING were paused before controller disposition, the candidate remains unheld, and no staging mutation occurred.

## Validation
- v2 acceptance suite: 10/10
- exact formerly failing shard after nested install: 70 suites / 513 tests
- targeted ESLint, Prettier, and `git diff --check`

Release notes intentionally disabled.

### DIFF
--- a/.github/workflows/release-bus-v2-preflight.yml
+++ b/.github/workflows/release-bus-v2-preflight.yml
@@ -186,9 +186,14 @@ jobs:
         with:
           node-version: '22'
           cache: npm
-          cache-dependency-path: package-lock.json
+          cache-dependency-path: |
+            package-lock.json
+            src/api-serverless/package-lock.json
       - name: Install frozen root dependencies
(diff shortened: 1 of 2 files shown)


GelatoGenesis:
GLM advisory validation is covered: the v2 acceptance test passed 10/10, and the exact formerly failing `tests-4` command passed 70 suites / 513 tests after the nested frozen install. The package-job cache suggestion is optional and does not affect correctness; this hotfix remains scoped to the live missing-dependency boundary.

PR #1796, opened 2026-07-23
fix: authorize exact v2 E2E manifest identity

GelatoGenesis:
## Summary
- give v2 authorization its own explicit artifact contract instead of inheriting v1 artifact-free rules
- require orchestration operations to be null/null, deployments to carry run+digest, and E2E to carry the exact manifest identity without an artifact run
- cover the exact staging E2E payload and reject missing/cross-operation manifest identities

## Live paused-beta evidence
Backend staging reached `STAGING_DEPLOYED` at exact SHA `f3e94829336a8627ffbe279fbb3c4f0d4f046419`. E2E run `29984983314` then stopped before tests because v1 validation required `artifact_digest:null`; the controller classified it CONTROL_PLANE and retained the exact manifest.

## Validation
- deploy validation + v2 operations suites: 35/35
- targeted ESLint, Prettier, and `git diff --check`

Release notes intentionally disabled.

### DIFF
--- a/src/api-serverless/src/deploy/deploy-validation.test.ts
+++ b/src/api-serverless/src/deploy/deploy-validation.test.ts
@@ -536,6 +536,29 @@ describe('Release Bus v2 validation', () => {
         operation_key: `rb2:${trainId}:deploy:staging:frontend:a1`
       }).error
     ).toBeDefined();
+    expect(
+      ReleaseBusV2AuthorizationBodySchema.validate({
+        ...authorization,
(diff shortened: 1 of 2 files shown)


GelatoGenesis:
Addressed the required review in the new head: shared authorization fields are factored once, E2E requires exactly five parsed key segments (`rb2`, train, `e2e`, environment, attempt), the nested collision is rejected, and ordinary artifact-backed deploys now have positive and missing-run tests. Targeted suites remain 35/35 green.

PR #1820, opened 2026-07-23
fix(release-bus-v2): retain retried fence runs

GelatoGenesis:
Fix the final staging fence so infrastructure retries preserve every exact workflow attempt run ID.

The fence reconstructs prior attempts from immutable operation keys and still fails closed if GitHub cannot prove one.

Focused validation: 45/45 v2 acceptance/reconciler tests, Prettier, targeted ESLint, diff check.

Release-note impact: none; internal Release Bus operational change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved staging validation to correctly ignore workflow runs from retried deployment and end-to-end attempts.
  * Prevented valid retried workflows from incorrectly blocking progression to staging validation.
  * Strengthened verification of ignored workflow run IDs during final-fence checks.


GelatoGenesis:
Superseded by #1821 with the identical GPG-signed diff plus the required DCO Signed-off-by trailer. No force-push was used.

PR #1831, opened 2026-07-24
Remove Simple Release Bus v1

GelatoGenesis:
## Summary
- remove all Simple Release Bus v1 API, controller, workflow, Step Functions, entity, configuration, and documentation surfaces
- preserve the existing v2 reconciler/function/schedule/invoke identities and replace the mixed cleaner with v2-only cleanup
- add fail-closed v2-only status/routing checks plus regression coverage for every retired v1 API method/path
- deliberately leave the historical v1 tables intact for a separately reviewed, snapshot-gated retirement after runtime drain

## Safe deployment order
1. Verify v1 executions/workflows/rows/leases are terminal.
2. Deploy api from the exact merged SHA to remove v1 routes and runtime configuration.
3. Deploy releaseBus from the same SHA to reduce the existing stack in place while preserving v2 identities.
4. Verify v2 and manual fallback live.
5. Delete v1-only repository variables and temporary v1 refs.
6. In a follow-up PR, snapshot production and retire the historical v1 tables only after this runtime deployment is proven.

## Validation
- targeted release-bus and route-census Jest suites passed, including all 12 retired v1 API method/path pairs
- TypeScript no-emit and ESLint on changed TypeScript passed
- src/releaseBus bundle build passed
- immutable Phase 0 route baseline retained with an explicit accepted-retirement exception for four v1 GET routes
- git diff --check passed


GelatoGenesis:
Addressed the prior verification gates at exact head `655cec9e`:

- The destructive migration no longer exists in this PR (`git diff origin/main...HEAD -- migrations` is empty). Historical v1 tables deliberately remain intact until a separate snapshot-gated follow-up after runtime/infrastructure retirement is live and proven.
- Added exact regression assertions that `serverless.yaml` wires `v2Reconciler` to `index.v2ReconcilerHandler` / `releaseBusV2Reconciler` and `cleaner` to `index.cleanerHandler` / `releaseBusCleaner`, while no starter, worker, state machine, or v1 mode remains. Targeted tests and ESLint pass.
- Repo-wide scans confirm there is no surviving v1 `RELEASE_BUS_MODE` consumer, service/repository claimant, starter/worker manifest entry, or unversioned v1 API/workflow caller; the only `RELEASE_BUS_MODE` occurrence is the intentional environment cleanup in deploy generation.
- Manual fallback remains intentionally available only after the audited fast-OFF route; active v2 mode is blocked at `/deploy/ui/dispatch`.

PR #1835, opened 2026-07-24
Fix V2 qualification immutable workflow refs

punk6529:
## Issue

A production-qualification train can reuse already prepared immutable artifacts without running repository preparation for the qualification train itself. In that path, no qualification-scoped frontend ref is created. The deploy can finish from the artifact, but manifest-bound staging E2E then fails closed because it cannot check out a ref resolving to the exact deployed frontend SHA, pausing Release Bus v2 globally.

## Fix

Bind the qualification train's immutable frontend workflow ref before it starts the staging idle handshake or acquires the staging environment lock.

## Notable changes

- Reassert the immutable frontend ref invariant at the staging/qualification lane boundary.
- Fail before staging ownership when the exact ref cannot be created idempotently.
- Cover successful qualification binding and the fail-before-lock path in the offline acceptance harness.

## Validation

- `npm test -- src/releaseBusV2/release-bus-v2.acceptance.test.ts --runInBand` — 48/48 passed
- `npm run lint` — passed under Git Bash on Windows
- `tsc -p tsconfig.json --noEmit` — passed after installing the API workspace dependencies
- `git diff --check` — passed

## Risk

Low and isolated to Release Bus v2 orchestration. Existing staging and multi-candidate refs are rechecked idempotently; a conflicting ref fails before the environment lock or any deployment mutation.

## Review notes

Please focus on the placement of the ref invariant before the idle snapshot and lock acquisition. No architecture documentation update is required because this restores an existing orchestration invariant and does not change the deployed system shape.

Deployment after merge: redeploy the `releaseBus` Lambda only, then verify locks/workflows are idle before resuming the global Release Bus v2 control.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved exact production qualification by binding the frontend workflow to an immutable release reference before staging proceeds.
  - Added validation to prevent progression when the required frontend revision is unavailable.
  - Improved failure handling so staging locks are not acquired when immutable reference creation fails.
  - Added regression coverage for immutable reference conflicts and related qualification behavior.

### DIFF
--- a/src/releaseBusV2/release-bus-v2.acceptance.test.ts
+++ b/src/releaseBusV2/release-bus-v2.acceptance.test.ts
@@ -9,6 +9,7 @@ const mockHasActiveStagingRun = jest.fn();
 const mockHasStagingRunSince = jest.fn();
 const mockHasActiveProductionRun = jest.fn();
 const mockFindWorkflowRun = jest.fn();
+const mockImmutableRefs = new Map<string, string>();
 
 jest.mock('@/releaseBusV2/release-bus-v2.operations', () => ({
   releaseBusV2Operations: {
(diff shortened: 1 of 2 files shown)


punk6529:
@coderabbitai review

punk6529:
Addressed the retry-coverage request in a0fedc54. The production GitHub App implementation is already idempotent: createRef treats HTTP 422 as success only when resolveRef returns the same exact SHA, and rejects a mismatched SHA; this is covered by the existing 'is idempotent only when a racing ref resolves to the exact SHA' unit test. The acceptance harness now models that exact-SHA behavior and drives PREPARED -> WAITING_FOR_ENVIRONMENT -> retry -> DEPLOYING, asserting both binds use the same qualification ref/SHA. The immutable ref is intentionally retained across waits; frontend qualification-branch creation triggers only the read-only all-branch secret scan, while deploy/E2E workflows are workflow_dispatch-only. The bare invariant error is classified by the reconciler's outer catch as CONTROL_PLANE via failTrain. Local focused suite is 49/49, focused lint passed, TypeScript compile passed, and diff check passed.

PR #1848, opened 2026-07-27
Keep staging refs aligned with deployed runtime

GelatoGenesis:
## Issue

- Managed Release Bus v2 staging could deploy and validate an exact cumulative runtime without advancing the affected repositories' shared `1a-staging` refs to that same immutable release.
- That allowed branch, runtime, manifest, and E2E identity to diverge, with ambiguous recovery after ref drift or a crash around mutation.

## Fix

- Compose each affected cumulative staging release as a forward-only immutable commit whose first parent is the recorded `1a-staging` SHA and whose tree is the dependency-closed candidate composition.
- Under the staging lock, CAS-advance only affected `1a-staging` refs before deployment, and require exact ref/runtime/manifest/E2E parity before validation.
- Persist ref advances as durable operations so a crash after GitHub accepts the CAS self-drains without repeating or guessing the mutation.

## Changes

- Add `release_parent_sha` to backend/frontend composition dispatch and validate reused immutable branches against the exact first parent.
- Remove carry-forward-only candidates from the deployment set while retaining them in cumulative composition.
- Record exact staging-ref identity in manifests and refuse manifest creation or E2E after drift.
- Detect manual/shared-ref drift before new claims, preserve the last validated manifest and candidate evidence, and pause only `STAGING`.
- Roll back forward: compose immutable restore commits from the failed release, CAS the affected refs, redeploy, and require rollback E2E before releasing staging.
- Preserve candidate identity when restoring historical manifests created before `candidate_id` was stored.
- Document the state-machine and zero-downtime ordering in the durable operations guide and architecture reference.

## Validation

- `npm run lint`
- `npm run build` — 292 suites and 2,679 tests passed after rebasing onto current `main`
- Focused Release Bus v2 suite — 87 tests passed
- `npm run tsc -- --noEmit`
- Latest Sonar follow-up refactor — ESLint, 80 focused tests, TypeScript, and diff check passed
- Composition workflow YAML parse
- `git diff --check`
- Clean merge-tree proof with candidate-evidence production PR #1847; its production fast-path guard and this staging-parent guard compose without conflict

## Risk

- Level: High
- Why: this changes the serialized shared-staging control plane and recovery behavior, although all ref mutations are non-force CAS operations and validation remains fail-closed.
- Rollback: pause `STAGING`; if needed deliberately switch v2 `OFF`, perform the documented serialized manual restore, and redeploy the prior `releaseBus` version. Never infer runtime identity or rewind a shared ref.

## Deployment

1. Merge the frontend companion first so its compose workflow accepts `release_parent_sha`.
2. Merge this PR so the backend compose workflow has the same contract.
3. Run the fresh release-bus status helper and require no active staging train or staging mutation/E2E.
4. Deploy only the backend `releaseBus` Lambda. No DB migration or API deployment is required.
5. Treat this as an internal operational release and use the release-note opt-out.

No staging or production mutation is part of this PR phase.

## Review Notes

- Frontend companion: 6529-Collections/6529seize-frontend#3484.
- Focus on compose → fence/lock → CAS → deploy callback → manifest → E2E sequencing, partial paired-CAS recovery, and forward-only rollback.
- This intentionally does not restore exact-combined-manifest production qualification. Candidate-evidence production PR #1847 remains the production policy and merge-tree combines cleanly.
- No Reset Staging control is added.

### DIFF
--- a/.github/workflows/release-bus-v2-compose.yml
+++ b/.github/workflows/release-bus-v2-compose.yml
@@ -10,6 +10,7 @@ on:
       expected_sha: { type: string, required: true }
       candidate_shas: { type: string, required: true }
       release_branch: { type: string, required: true }
+      release_parent_sha: { type: string, required: false, default: '' }
 
 permissions:
   contents: read
@@ -28,6 +29,7 @@ jobs:
           EXPECTED_SHA: ${{ inputs.expected_sha }}
           OPERATION_KEY: ${{ inputs.operation_key }}
           RELEASE_BRANCH: ${{ inputs.release_branch }}
+          RELEASE_PARENT_SHA: ${{ inputs.release_parent_sha }}
           TRAIN_ID: ${{ inputs.release_train_id }}
         run: |
           set -euo pipefail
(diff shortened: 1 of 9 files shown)


GelatoGenesis:
Superseded by #1849 on a fresh branch so the missing DCO trailer could be corrected without force-pushing shared history.

PR #1856, opened 2026-07-27
fix: reduce drop write lock contention

punk6529:
## Issue
Production `POST /api/drops` requests intermittently reached the API/Lambda timeout while users pasted and submitted text. Runtime evidence showed a row-lock convoy on the shared per-wave metrics rows; the database remained available and Lambda was not throttled.

## Fix
Move the `wave_metrics` and `wave_dropper_metrics` increments to the transaction tail for new drops. This keeps the same atomic transaction semantics while sharply reducing how long contended metric rows remain locked. Edits no longer increment current-drop counters, and insert/delete paths acquire the shared rows in one deterministic order.

Direct `insertDrop` callers keep the existing immediate metric behavior unless they explicitly opt into deferral.

## Notable changes
- extract a typed and timed `applyInsertedDropMetricsDelta` operation
- opt the full drop persistence transaction into deferred metrics updates
- apply the deferred increment only for new drops, not edits
- use deterministic wave-before-dropper metric lock ordering for insert and delete paths
- add regression coverage for deferred, direct-call, and update behavior

## Validation
- `npm run lint` — pass
- `npm run tsc -- --noEmit` — pass
- focused Jest suites for drop DB, create/update use case, and drop creation API — 83/83 pass
- full local build test phase — 2,692/2,721 pass; three unrelated Release Bus suites fail only in the Windows child-process harness because tool PATH/ref text is not normalized there. Linux PR CI is authoritative for those suites.

## Risk and rollback
Medium-low. New-drop metric writes remain in the same database transaction and use the same deltas; their ordering changes and edits stop over-counting current-drop totals. Roll back by reverting these commits and redeploying the API service.

## Deployment
Backend `api` service only. No schema migration, frontend change, or user-facing documentation change.

## Review notes
Please focus on transaction-tail ordering, unchanged rollback/atomicity for new drops, deterministic metric lock order, and compatibility for direct `insertDrop` callers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected drop metrics so they are recorded only for newly created drops, not when existing drops are edited.
  * Improved deferred metrics handling to prevent duplicate or premature counter updates.
* **Tests**
  * Added coverage for editing drops without incrementing insertion metrics.
  * Added tests verifying both deferred and immediate metrics updates during drop insertion.

### DIFF
--- a/src/drops/create-or-update-drop-use-case.test.ts
+++ b/src/drops/create-or-update-drop-use-case.test.ts
@@ -164,6 +164,64 @@ describe('CreateOrUpdateDropUseCase', () => {
     };
   }
 
+  it('does not increment inserted-drop metrics when editing a drop', async () => {
+    const dropsDb = {
+      findDropById: jest.fn().mockResolvedValue({
+        id: 'drop-1',
+        wave_id: 'wave-1',
+        author_id: 'author-profile',
+        drop_type: DropType.CHAT,
+        created_at: Date.now() - 1000,
+        updated_at: Date.now(),
+        serial_no: 1
+      }),
+      getDropGroupMentions: jest.fn().mockResolvedValue([]),
+      applyInsertedDropMetricsDelta: jest.fn().mockResolvedValue(undefined)
+    };
+    const wavesApiDb = {
+      findById: jest.fn().mockResolvedValue({
+        ...createSlowModeWave(),
+        type: WaveType.CHAT
+      })
+    };
+    const deleteDropUseCase = {
+      execute: jest.fn().mockResolvedValue(undefined)
+    };
+    const artCurationTokenWatchService = {
+      registerDrop: jest.fn().mockResolvedValue(undefined)
+    };
+    const useCase = createUseCaseWithMocks({
+      dropsDb,
+      wavesApiDb,
+      deleteDropUseCase,
+      artCurationTokenWatchService
+    });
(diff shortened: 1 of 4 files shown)


punk6529:
Addressed the DB/general findings in `5b4eca0d`:

- gated the deferred metrics delta to new drops, with an update-path regression test proving edits perform no inserted-drop increment;
- made delete acquire `wave_metrics` before `wave_dropper_metrics`, matching insert order deterministically;
- documented the intentional lock order and re-ran lint, TypeScript, and the three focused suites (83/83).

Historical note: the pre-PR implementation also incremented these counters during update because both branches called `insertDrop`; the new guard corrects that existing over-count rather than preserving it. I left the timer nice-to-have unchanged because the direct-call compatibility signature has no timer context; the incident path passes its timer to the extracted tail operation.

punk6529:
The transaction-context advisory is covered in 7141a92b: both deferred metric queries now explicitly assert the same wrapped connection used by the drop transaction. The focused DB suite remains 15/15. The failure-path advisory is already guaranteed by control flow: the tail call is reached only after insertAllDropComponents resolves, and any sibling rejection propagates before it; rollback is owned by the surrounding transaction boundary.

punk6529:
@coderabbitai review

punk6529:
@coderabbitai full review

PR #1859, opened 2026-07-28
Add surgical production TDH repair for Meme 473

prxt6529:
## Issue
- Production recorded transaction `0xd6af5074fcdb210a5cfc2e85ad9f512e684271cc9158454cbb123d134efef30f` for Meme #473 with `token_count = 2`, although the on-chain transfer quantity is `1`.
- That doubled transfer left the two raw owner rows and both affected consolidated-owner rows one edition out, producing the observed SZN14 TDH discrepancy.

## Fix
- Add a one-shot `customReplayLoop` repair that validates the complete known-bad production state and defaults to dry-run.
- Apply the five exact corrections atomically only when every target row and value matches the expected pre-repair snapshot.

## Changes
- Target the exact transaction hash, block, contract, token, sender, recipient, and consolidation keys.
- Lock and verify the transaction row, sender/recipient owner rows, and sender/recipient consolidated-owner rows.
- Require exactly one affected row for each update and verify the complete repaired state before commit.
- Return safely when the database is already in the repaired state; abort on missing, duplicate, partial, or unexpected state.
- Skip Redis initialization and local migration execution for this surgical replay path.
- No schema, entity, API contract, dependency, or architecture changes.

## Validation
- Focused replay-state tests pass: 6 tests covering bad, repaired, partial, missing, duplicate, and unexpected states.
- ESLint passes for all four changed files; the repo-wide lint command also completed without errors.
- The full TypeScript command was attempted but the available local dependency installation is incomplete for current `main` and reports unrelated missing packages/types. The focused test compiles and exercises the repair module through `ts-jest`.
- The default Jest bootstrap could not start its MySQL testcontainer because Docker is unavailable in this environment, so no production-connected or write-path integration test was run.

## Risk
- Level: High
- Why: This is intentionally capable of mutating five production rows that feed ownership and TDH.
- Mitigation: dry-run is the default; exact preconditions, pessimistic locks, one database transaction, exact affected-row assertions, post-write verification, and idempotency prevent a broad or partial write.
- Rollback: do not apply unless the dry-run reports the exact known-bad snapshot. If post-write verification fails, the transaction rolls back automatically.

## Deployment
- None. This one-shot repair is run directly from the reviewed branch; no backend service deployment is required.
- Production runbook:

  ```bash
  CUSTOM_REPLAY_APPLY=false npm run backend:local -- customReplayLoop
  CUSTOM_REPLAY_APPLY=true npm run backend:local -- customReplayLoop
  CUSTOM_REPLAY_APPLY=false npm run backend:local -- customReplayLoop
  ```

- Expected results are `DRY_RUN_VERIFIED`, then `REPAIRED`, then `ALREADY_REPAIRED`.
- Run the apply step only after the first dry-run output is reviewed and explicit production authorization is given. Then verify the production TDH endpoint and the five repaired rows.

## Review Notes
- Please focus on the hard-coded target identifiers, expected before/after balances, and the all-or-nothing transaction behavior.
- This PR does not address the underlying duplicate-ingestion cause; it is limited to repairing the current production data.

### DIFF
--- a/src/backend.ts
+++ b/src/backend.ts
@@ -84,11 +84,18 @@ async function runRequestedLoopIfPresent() {
 async function start() {
   logger.info(`[CONFIG ${process.env.NODE_ENV}] [EXECUTING START SCRIPT...]`);
 
-  await dbMigrationsLoop.handler(
-    undefined as any,
-    undefined as any,
-    undefined as any
-  );
+  const requestedLoopName = getRequestedLoopName();
+  if (requestedLoopName === 'customReplayLoop') {
+    logger.info(
+      `[SKIPPING DB MIGRATIONS FOR SURGICAL LOOP] [${requestedLoopName}]`
+    );
+  } else {
+    await dbMigrationsLoop.handler(
+      undefined as any,
+      undefined as any,
+      undefined as any
+    );
+  }
 
   await runRequestedLoopIfPresent();
 
--- a/src/customReplayLoop/index.ts
+++ b/src/customReplayLoop/index.ts
@@ -1,6 +1,9 @@
 import { doInDbContext } from '../secrets';
 import { Logger } from '../logging';
 import * as sentryContext from '../sentry.context';
+import { ConsolidatedNFTOwner, NFTOwner } from '../entities/INFTOwner';
+import { Transaction } from '../entities/ITransaction';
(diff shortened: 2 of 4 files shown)


prxt6529:
Production data repair completed and verified. Closing this one-shot repair PR; the underlying ingestion fix will be handled in a separate PR.

PR #1865, opened 2026-07-28
Fix Release Bus production replan coalescing

GelatoGenesis:
## Summary

- transactionally preserve exact explicit production intents when a production train is safely replanned before irreversible mutation
- claim a new audited replacement from every currently eligible explicit selection, with fresh exact-head, staging evidence, base, ownership, revocation, and dependency checks
- freeze the original immutable set after any successful main advance, production deploy dispatch, or production E2E creation
- serialize mark-ready and replan/claim races with the scheduler fence and retain exact omission reasons without inferring intent from staging

## Why

A replacement train currently preserves only the source train's original production selection identity. A later compatible explicit selection can therefore wait behind the full replacement pipeline even though it was committed before the replacement claim. This change coalesces all currently eligible explicit intents only at the safe pre-mutation replacement boundary and never broadens an active train.

## Safety and compatibility

- no migration or API contract change
- no force updates; existing non-force CAS main advancement is unchanged
- production serialization, idempotency keys, retry bounds, deploy/E2E exactly-once behavior, and manual fallback remain unchanged
- selection holds the scheduler lease across authoritative commit/rollback; replan defers nonterminal dispatched composition
- frozen trains retain the production lease only while dispatched work drains and release it after terminal work
- omitted or invalid intents are re-read under lock and stay auditable with exact source selection/train mapping and reason
- replacement scans fail closed on truncation or a moved production base, with operator recovery documented
- rollout order: `api` first, then `releaseBus`; production release-note opt-out applies because this is an internal operational change

## Verification

- `npm run build` (296 suites, 2,797 tests, TypeScript)
- focused Release Bus gate (4 suites, 148 tests)
- `npm run lint`
- `npm run format:check`
- `git diff --check`

## Coordination

Rebased onto backend main `8c96a72860daa69a669c059d5e5a54e646a37875` after #1864 merged. The staging-recovery ownership/supersession and repair behavior is retained. Registration and rollout will wait for that production deployment and every Release Bus lock/workflow to drain.

## Summary by CodeRabbit

- **New Features**
  - Added safe production replanning when the production base changes before irreversible deployment work.
  - Preserves eligible production intents for a replacement train while keeping dependency closure and ownership intact.
  - Freezes the original train once deployment progress becomes irreversible.

- **Bug Fixes**
  - Added fail-closed handling for bounded production-intent scans and concurrent scheduling races.
  - Improved validation of production evidence, dependencies, and current production state.

- **Documentation**
[...]

PR #1866, opened 2026-07-28
Fix Release Bus production replan coalescing

GelatoGenesis:
## Summary

- transactionally preserve exact explicit production intents when a production train is safely replanned before irreversible mutation
- claim a new audited replacement from every currently eligible explicit selection, with fresh exact-head, staging evidence, base, ownership, revocation, and dependency checks
- freeze the original immutable set after any successful main advance, production deploy dispatch, or production E2E creation
- serialize mark-ready and replan/claim races with a committed scheduler row lease and retain current exact omission reasons without inferring intent from staging
- bound every Release Bus GitHub App request so external latency cannot strand scheduler/candidate locks

## Why

A replacement train currently preserves only the source train original production selection identity. A later compatible explicit selection can therefore wait behind the full replacement pipeline even though it was committed before the replacement claim. This change coalesces all currently eligible explicit intents only at the safe pre-mutation replacement boundary and never broadens an active train.

## Safety and compatibility

- no migration or API contract change
- no force updates; existing non-force CAS main advancement is unchanged
- production serialization, idempotency keys, retry bounds, deploy/E2E exactly-once behavior, and manual fallback remain unchanged
- selection and replan both hold the same scheduler row lease across authoritative commit/rollback; the database integration test proves the lease excludes a concurrent transactional claimant
- transient production-base infrastructure failures roll back and retry on a later scheduler tick without mutating intent or pausing the lane
- replan defers all nonterminal dispatched composition, including retry wait
- frozen trains retain the production lease only while dispatched work drains and release it after terminal work
- omitted or invalid intents are re-read under lock and stay auditable with current exact reason plus source selection/train provenance; dependent omissions carry their prerequisite root cause
- replacement scans fail closed on truncation or a moved production base, with counts that distinguish scanned from eligible rows
- GitHub request options exclude caller signals, making the helper-owned timeout the single authoritative abort source and classification
- rollout order: `api` first, then `releaseBus`; production release-note opt-out applies because this is an internal operational change

## Verification

- `npm run build` (296 suites, 2,805 tests, TypeScript)
- focused Release Bus gate (5 suites, 184 tests, including the real database scheduler-lock race proof)
- review-fix gate (2 suites, 86 tests; final GitHub-app suite 25 tests plus TypeScript)
- `npm run lint`
- `npm run format:check`
- `git diff --check`
- exact-head GitHub CI run `30368586589`

## Review disposition

- 6529bot follow-up on final head `e2f02046b207815e46f121dda21f81d40399dedf`: no new findings
- substantive CodeRabbit full review completed across all nine files; both trivial findings are addressed by `95a9a88a7c932ccbea6b74da381fb950e9e6e09c` and `e2f02046b207815e46f121dda21f81d40399dedf`

## Coordination

Based on backend main `8c96a72860daa69a669c059d5e5a54e646a37875` after #1864 merged and reached production. The staging-recovery ownership, supersession, and repair behavior is retained. Runtime rollout will wait for every active Release Bus train, workflow, and lock to drain.

## Supersession

Supersedes #1865 without rewriting its published branch. This branch started from the same reviewed final tree as #1865 head `f4b8381650f74b9caedfe2bd3823a2135d1b38b7`, recorded as a GPG-signed and DCO-signed-off commit, then added review-driven hardening and the two final review dispositions as separately signed and signed-off commits. No published history was rewritten.

### DIFF
--- a/ops/deployment-bus/simple-release-bus-v2.md
+++ b/ops/deployment-bus/simple-release-bus-v2.md
@@ -138,6 +138,34 @@ candidate/repository/PR/head identity and successful production E2E. Omitted
 unrelated candidates keep their staging evidence and separate production
 intent.
 
+If either production base moves before irreversible production mutation, the
+old train is cancelled without changing its immutable membership and its exact
+explicit intents move to `WAITING_FOR_PRODUCTION_REPLAN`. The next scheduler
+transaction creates a new replacement selection and train from every currently
+eligible explicit production intent, including compatible selections committed
+after the old train was claimed. It locks and rechecks exact heads, staging
+train/manifest/E2E evidence, dependencies, ownership, and both current
+production bases. Audit events map every included candidate and source
+selection/train to the replacement and retain every omitted intent with its
+exact reason. Revoked, superseded, moved, stale-evidence, dependency-incomplete,
+or concurrently owned candidates are never inferred from staging and are not
+silently included.
+
+If `PRODUCTION_REPLAN_INTENT_SCAN_FAILED_CLOSED` reports the 500-row scan cap,
+no replacement may claim. Wait for production ownership to drain, inspect every
+explicit ready/held intent, and use the authenticated revoke/cancel actions
+only for owner-authorized stale intents until the bounded scan is complete.
+Otherwise deploy a separately reviewed cap/pagination change. Never edit the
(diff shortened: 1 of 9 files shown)


GelatoGenesis:
@coderabbitai review

GelatoGenesis:
@coderabbitai full review

PR #1893, opened 2026-07-30
Fix failed baseline adoption operation cleanup

GelatoGenesis:
## What changed

- terminalize an exact baseline-adoption E2E operation when its train fails before dispatch reservation
- make the operator-only idempotent adoption endpoint repair an older failed intent left in that proven-undispatched state
- append deterministic audited recovery evidence
- refuse to cancel DISPATCHED/ambiguous operations

## Root cause

The failed baseline-adoption path marked the manifest and train FAILED and released the staging lock, but could leave the already-created E2E operation PENDING. Manual fallback correctly treats that nonterminal operation as a drain blocker.

## Impact

The same shared control-plane logic now cleans this state for staging frontend/backend baseline adoption, while preserving ambiguous workflow state for reconciliation. The fix is deployed through both production and staging API/control-plane runtimes before STAGING is enabled.

## Validation

CI-first per owner request. Added focused regressions for new failures, recovery of the existing state, and refusal to guess after dispatch reservation. Formatting and diff checks passed locally; broad suites are delegated to PR CI.

### DIFF
--- a/src/releaseBusV2/release-bus-v2.baseline-adoption.test.ts
+++ b/src/releaseBusV2/release-bus-v2.baseline-adoption.test.ts
@@ -1381,6 +1381,59 @@ describe('ReleaseBusV2BaselineAdoptionService', () => {
     );
   });
 
+  it('keeps failed intent reads stable when recovery races, then retries cleanup', async () => {
+    const context = harness();
+    await prepare(context);
+    await freeze(context);
+    Object.assign(context.repository.operations[0], {
+      status: 'PENDING',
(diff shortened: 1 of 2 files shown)


punk6529 (approved):
Reviewed against the live failed train/operation state. The fix only terminalizes an exact PENDING baseline-adoption E2E with no external reservation, preserves DISPATCHED/ambiguous operations, is idempotent, and adds audited recovery evidence plus regressions. Approved for the release-blocker recovery path.

PR #1897, opened 2026-08-02
Fix exact Stream development status answers

punk6529:
## Issue

The live Stream Help Bot path could mix the current daily development update with historical snapshot evidence, and Bedrock could then invent exact bytecode figures or launch checklist items. Root-anchored review evidence paths were also rejected by the shard validator.

## Fix

- route current Stream development questions to the single versioned \development_status\ record
- project its exact checked date, completed work, work in progress, and before-launch checklist into a bounded packet
- render those status facts deterministically while keeping Bedrock for ordinary Stream explanations
- accept validated Stream root-anchor evidence paths
- document current-status precedence in the runtime spec

## Validation

- 72 focused Help Bot tests pass
- repository TypeScript typecheck passes after installing root and API dependencies
- full source ESLint passes with zero warnings
- Prettier and \git diff --check\ pass

## Risk

Low and scoped to Stream Help Bot retrieval/rendering. Pinned snapshot questions keep the existing evidence path; only clearly current/development-status questions use the deterministic projection.

## Review notes

The regression fixture intentionally contains both the current 5,579-byte headroom fact and older 424/1,576-byte snapshot facts, proving that the current answer packet contains only the dated development record.

### DIFF
--- a/specs/6529-help-bot-runtime.md
+++ b/specs/6529-help-bot-runtime.md
@@ -171,7 +171,9 @@ items remaining before launch select only that record. Historical risk and
 readiness records describe the pinned review snapshot and must not be mixed into
 an answer about the current development update. The backend projects the exact
 status fields into the answer and bypasses Bedrock for those facts so figures
-and checklist items cannot be rewritten or invented.
+and checklist items cannot be rewritten or invented. Its provenance commit is
+the separately checked development commit, while the manifest source commit
+continues to identify the pinned review snapshot.
 
 ### 4.3 Future docs chunking and RAG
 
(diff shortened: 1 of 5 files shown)


punk6529:
Addressed the valid general-review finding in signed commit 1bb4bb28: contract-size matching now prefers a matching completed item but falls back to the full exact completed list, and an otherwise empty requested section falls back to the exact headline/work/launch projection. The regression test now uses alternate EIP-170 wording. Focused suite remains 72/72 green; TypeScript and ESLint are clean.

punk6529:
Addressed the valid GLM advisory on historical intent in signed commit e3c681d4: historical/pinned/reviewed wording excludes the daily development record, the daily record is selected by stable ID, and tests cover pinned snapshot wording, reviewed-version wording, current development status, and an unscoped headroom query. Focused suite is now 75/75 green; TypeScript and ESLint remain clean.

PR #1898, opened 2026-08-02
Preserve exact Stream status payload

punk6529:
## Issue

The first live staging retest after #1897 returned the correct launch checklist but rounded the exact 5,579-byte headroom to 2.5 KB. The projected development status plus the record's long summary crossed the generic per-record budget, so the formatter replaced structured fields with a truncated excerpt and the deterministic answer path could not engage.

## Fix

- give \development_status\ a dedicated minimal evidence payload containing only identity, canonical path, provenance commit, and the exact structured status fields
- remove evidence paths from the runtime status item projection because the public answer needs the exact item IDs/text only
- stress the regression fixture with a long narrative and assert the complete structured payload remains present and within the 2,300-character record limit

## Validation

- 75 focused Help Bot tests pass
- TypeScript typecheck passes
- ESLint, Prettier, and \git diff --check\ pass
- live staging evidence that exposed the issue: checklist exact, byte value incorrectly rounded; this PR targets that packet boundary directly

## Risk

Low. The special formatter applies only to the stable \status:latest-development\ record selected in #1897. Other Stream evidence keeps the generic bounded formatter.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved development-status information formatting for clearer, more compact evidence.
  * Removed unnecessary evidence details from development-status records.
  * Ensured long evidence summaries remain within supported size limits.

### DIFF
--- a/src/help-bot/help-bot-answerer.test.ts
+++ b/src/help-bot/help-bot-answerer.test.ts
@@ -322,6 +322,59 @@ describe('HelpBotAnswerer', () => {
     expect(renderer.renderAnswer).not.toHaveBeenCalled();
   });
 
+  it('fails closed without the LLM when an exact Stream status packet is oversized', async () => {
+    const renderer: HelpBotLlmRenderer = {
+      renderAnswer: jest.fn().mockResolvedValue('Invented status')
+    };
+    const streamKnowledgeSource: HelpBotStreamKnowledgeSource = {
+      findMatch: jest.fn().mockResolvedValue({
+        score: 400,
+        record: {
+          id: '6529-stream@2026-08-01.1',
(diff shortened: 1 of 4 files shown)


punk6529:
@coderabbitai review

punk6529:
Addressed the valid size-cap review in the latest signed commit: the full exact development payload is used when it fits; an oversized future structured status now emits a bounded \exactStatusUnavailable\ marker, and the answerer responds deterministically with the verified source link without invoking Bedrock. The test inflates \structured.summary\ (a field actually emitted), asserts the record cap and marker, and a separate answerer test proves the LLM is bypassed. Focused suite is 77/77 green; TypeScript and ESLint are clean.

PR #1930, opened 2026-08-17
Validate Wave privilege groups against View

GelatoGenesis:
## Summary

- enforce View-group containment for active Drop, Vote, Chat, and Admin scopes on Wave create/update
- evaluate all explicit group relationships in one batched SQL anti-join, with public/same-group fast paths
- expose a lightweight validation endpoint for the frontend and preserve the invariant across published group replacements
- intersect runtime privilege flags and management checks with View eligibility for legacy or dynamically drifted groups

## Testing

- `./bin/6529 run lint`
- `./bin/6529 run build:ci`
- 48 focused Jest tests passed without the container-backed global setup
- added container-backed MySQL coverage for the composed membership anti-join (local full run unavailable because this host has no container runtime)

## Rollout

Deploy the backend `api` unit before the dependent frontend PR. No schema migration is required; the API change is additive and backward compatible.

PR #1936, opened 2026-08-18
Validate Wave privilege groups against View

GelatoGenesis:
## Issue

- Waves could be created or edited with Drop, Vote, Chat, or Admin access granted to people who could not view the Wave.

## Fix

- Enforce View-group containment server-side for Wave creation, Wave updates, and published group replacements.
- Evaluate explicit group relationships with a batched SQL anti-join plus public and same-group fast paths.

## Changes

- Add an additive validation endpoint for frontend preflight checks.
- Intersect runtime privilege and management checks with View eligibility for legacy or dynamically drifted groups.
- Add generated API types, containment tests, SQL membership coverage, and architecture documentation.

## Validation

- `./bin/6529 run check:changed`
- 127 related Jest suites passed (1,227 tests).
- Equivalent superseded-PR CI passed the backend build, SonarCloud, and Snyk checks.
- OpenAPI output matches the dependent frontend candidate.

## Risk

- Level: Medium
- Why: this tightens Wave authorization and adds an endpoint, but makes no schema change and keeps the API additive/backward compatible.
- Rollback: revert the single signed commit and redeploy the `api` unit.

## Review Notes

- Deployable unit: `api` only. Deploy it before the dependent frontend candidate.
- This clean signed PR replaces #1930; its code and resolved review fixes are preserved without rewriting shared history.


GelatoGenesis:
Addressed the latest 6529bot findings in `9485e5ab5`:

- hardened membership-SQL namespacing: validates namespaces, preserves double-colon tokens, follows MySQL backtick semantics, fails closed on open quotes, and documents the trusted-generator/bound-value invariant
- added parser tests with quoted colons, double-colon text, and colon-containing bound values
- load only active Wave scopes, made invalid-role response ordering explicit, and added chat-disabled View-replacement coverage
- documented that optional auth consumes `req.user` populated by `maybeAuthenticatedUser`

Existing safeguards verified: the app-wide rate-limiting middleware wraps the generated router for anonymous and authenticated callers; optional-auth handlers conventionally call `getAuthenticationContext` after that middleware; the preview service test already proves unavailable ids use the generic non-disclosing error; and concurrent group replacements serialize on the same affected Wave row locks inside the replacement transaction.

GelatoGenesis:
Follow-up `09cfd6e08` clears the new Sonar findings from the prior head: the membership scanner is split into focused quote/parameter readers, no longer mutates a `for`-loop index, and retains an own-property check compatible with the repository TypeScript target. The optional-chain readability finding is also fixed. Full lint and `check:changed` remain green (127 suites / 1,230 tests).

